PR #3, opened 2016-09-30
Warn against combining yield and return-with-value

### LINKED ISSUE #2: Warn against using yield and return x
[Python 3 allows mixing of return-with-value and yield in the same function](https://bugs.python.org/issue28310). I got bitten by this behavior multiple times, so I want to write a flake8 plugin for it. I wonder if I should add the lint to this package instead of making a separate one. What do you think?

untitaker:
Fix #2

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,6 +74,10 @@ builtin. For Python 2 compatibility, use ``six.next()``.
 and is removed in Python 3. Use ``str(e)`` to access the user-readable
 message. Use ``e.args`` to access arguments passed to the exception.
 
+**B307**: Using ``return x`` in a generator function is syntactically invalid
+in Python 2, but in Python 3 ``return x`` in a generator function means ``raise
+StopIteration(x)``.
+
 
 Tests
 -----
--- a/README.rst
+++ b/README.rst
@@ -74,9 +74,19 @@ builtin. For Python 2 compatibility, use ``six.next()``.
 and is removed in Python 3. Use ``str(e)`` to access the user-readable
 message. Use ``e.args`` to access arguments passed to the exception.
 
-**B307**: Using ``return x`` in a generator function is syntactically invalid
+
+Opinionated warnings
+~~~~~~~~~~~~~~~~~~~~
+
+Those warnings are disabled by default because the opinions they come from are
+controversial.
+
+**B901**: Using ``return x`` in a generator function is syntactically invalid
 in Python 2, but in Python 3 ``return x`` in a generator function means ``raise
-StopIteration(x)``.
+StopIteration(x)``. Users that come from Python 2 may be used to Python 2
+rejecting code that mixes both, so this can lead to situations where it looks
+like ``return x`` makes the function generate an empty sequence. Use the more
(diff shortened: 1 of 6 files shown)


untitaker:
I know that I didn't wait for a response on #2. This is just a draft, feel free to review or discard.

untitaker:
@ambv This is ready for review

README.rst
```diff
@@ -75,6 +75,20 @@ and is removed in Python 3. Use ``str(e)`` to access the user-readable
...
+~~~~~~~~~~~~~~~~~~~~
+
+Those warnings are disabled by default because the opinions they come from are
+controversial.
+
+**B901**: Using ``return x`` in a generator function is syntactically invalid
+in Python 2, but in Python 3 ``return x`` in a generator function means ``raise
+StopIteration(x)``. Users that come from Python 2 may be used to Python 2
+rejecting code that mixes both, so this can lead to situations where it looks
+like ``return x`` makes the function generate an empty sequence. Use the more
```
ambv: The suggestion to use `raise StopIteration(x)` as an equivalent is invalid. See PEP 479. StopIterations that come out of generators are not transformed into RuntimeErrors, see specifically:
https://www.python.org/dev/peps/pep-0479/#writing-backwards-and-forwards-compatible-code

Since valid users of this functionality are most likely users of coroutines, a suggestion to use native coroutines here would be enough.

Add a check in your unit test that confirms that native coroutines (`async def` with `await` and `return`) don't raise B901 and we're good.

bugbear.py
```diff
@@ -229,3 +254,13 @@ def compose_call_path(self, node):
             "to the exception.",
     type=BugBearChecker,
 )
+
+B901 = partial(
+    error,
+    message=("B901: Using ``yield`` together with ``return x``. Use ``raise "
+             "StopIteration(x)`` instead of ``return x`` if this was "
+             "intentional."),
```
ambv: Again, because of PEP 479, I'd :

"B901: Using ``yield`` together with ``return x`` is confusing to the reader. Use native ``async def`` coroutines or put a ``# noqa`` comment on this line if this was intentional."

README.rst
untitaker: Interesting. This code prints `1` under Python 3.5, where this PEP was supposedly implemented:

    def generator():
        yield 1
        raise StopIteration()
        yield 2
        yield 3

    for x in generator():
        print(x)

ambv: Does it also do that if you add `from __future__ import generator_stop`?

bugbear.py
untitaker: The mention of async def coroutines is a good idea, but I don't consider `return x  # noqa` to be more readable than `return x`. Readability is all this check is about!

ambv: Well, if the user *really* uses `return x` correctly and intentionally in a generator, how is she supposed to let the linter know?

If you add `# noqa` to the line, you are letting both the linter and the human reader know that you are in fact doing this on purpose.

untitaker: I was assuming that `raise StopIteration(x)` is a viable alternative to `return x`, so I guess recommending `# noqa` is the only option.

PR #4, opened 2016-10-04
Fix bug in B901 check wrt closures

untitaker:
Just found this bug while running bugbear against my own code. Really sorry for not checking sooner!


ambv:
I will never forgive you but I will likely forget 😆 

Seriously though, don't beat yourself up about this. You noticed this real fast and fixed it. That's what matters :)

untitaker:
Thanks, could you release a new version?

ambv:
Done.

PR #12, opened 2017-02-25
B902 now deals with metaclasses

### LINKED ISSUE #10: B902 requires first parameter named self for __init__ of metaclass
I have a metaclass defined like this:

    class StorageMeta(type):
        def __init__(cls, name, bases, d):

Naturally B902 flagged the first parameter of `__init__`, but I'm not sure if it's intended in the case of metaclasses. I can imagine arguments for and against this behavior.

untitaker:
Fix #10

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,9 +107,10 @@ Users coming from Python 2 may expect the old behavior which might lead
 to bugs.  Use native ``async def`` coroutines or mark intentional
 ``return x`` usage with ``# noqa`` on the same line.
 
-**B902**: Invalid first argument used for method. Use ``self`` for
-instance methods, and `cls` for class methods (which includes `__new__`
-and `__init_subclass__`).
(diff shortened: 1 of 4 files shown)


ambv:
Thanks! I'll do some minor changes on this, for instance to allow class methods on types use a different first argument name (the official docs use `metacls`, see the OrderedClass example here: https://www.python.org/dev/peps/pep-3115/).

@untitaker, why did you end up using `getattr()` and not `x.id` directly?

untitaker:
@ambv I wasn't entirely sure if there are situations where one would put anything else than an identifier here. For example one could write code like this:

    class Foo("bar"):
        pass

and while it is complete nonsense it shouldn't crash the linter like that, I think.

untitaker:
Also I probably would've figured out how to configure the first arg if you asked me to. :)

ambv:
No worries, man, thanks for taking your time with it already :)

PR #13, opened 2017-02-28
Proposed test: b903, data classes should be namedtuples or slots

mjpieters:
This tests for

```python
class Foo:
    def __init__(self, bar, spam, eggs):
        self.bar = bar
        self.spam = spam
        self.eggs = eggs
```

resulting in

> B903 Data class should either be immutable or use \_\_slots__ to save memory. Use collections.namedtuple to generate an immutable class, or enumerate the attributes in a \_\_slot__ declaration in the class to leave attributes mutable.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -380,14 +380,21 @@ class BugBearVisitor(ast.NodeVisitor):
             )
 
     def check_for_b903(self, node):
-        if (len(node.body) != 1 or
-                not isinstance(node.body[0], ast.FunctionDef) or
-                node.body[0].name != '__init__'):
+        body = node.body[:]
+        if body and isinstance(body[0], ast.Expr) and isinstance(body[0].value, ast.Str):
+            # Ignore the docstring
+            body = body[1:]
+
+        if (
+            len(body) != 1 or
+            not isinstance(body[0], ast.FunctionDef) or
+            body[0].name != '__init__'
+        ):
             # only classes with *just* an __init__ method are interesting
             return
 
         # all the __init__ function does is a series of assignments to attributes
-        for stmt in node.body[0].body:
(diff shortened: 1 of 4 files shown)


untitaker:
That's interesting, though I would list the recommendation for using `__slots__` first (since it is more likely to be applicable than using namedtuple)

mjpieters:
Why is it more likely? The instances of the specific case this was inspired by were never mutated; I'd err on the side of immutability.

It is far easier to discover that your `namedtuple` instances should really be mutable than it is to discover that your `__slots__` class instances should not be mutable. And for most cases you really want to use `namedtupinstance._replace(fieldname=newvalue)` anyway.

untitaker:
> I'd err on the side of immutability.

I wouldn't since we have no way of detecting that, but I guess it's not an important thing to change.

bugbear.py
```diff
@@ -375,6 +379,26 @@ def check_for_b902(self, node):
                 )
             )
 
+    def check_for_b903(self, node):
+        if (len(node.body) != 1 or
+                not isinstance(node.body[0], ast.FunctionDef) or
+                node.body[0].name != '__init__'):
```
mjpieters: Ah, the eternal `if` over-indent discussion! I'll comply with your preference and I'll add support for `__doc__` too.

bugbear.py
```diff
@@ -380,14 +380,21 @@ def check_for_b902(self, node):
             )
 
     def check_for_b903(self, node):
-        if (len(node.body) != 1 or
-                not isinstance(node.body[0], ast.FunctionDef) or
-                node.body[0].name != '__init__'):
+        body = node.body[:]
```
ambv: Why copy it? You don't mutate it.

PR #16, opened 2017-03-02
Fix typos

jwilk:
Found using [mwic](http://jwilk.net/software/mwic).


untitaker:
Neat tool, thanks for "pitching" it! I've been looking for something like it.

ambv:
Thanks! ✨ 🍰 ✨

PR #21, opened 2017-10-15
Include license file in the generated wheel packages

jdufresne:
The wheel package format supports including the license file. This is done using the `[metadata]` section in the `setup.cfg` file. For additional information on this feature, see:

https://wheel.readthedocs.io/en/stable/index.html#including-the-license-in-the-generated-wheel-file

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,3 +8,6 @@ ignore = E302, E501
 max-line-length = 80
 max-complexity = 12
 select = B,C,E,F,W,B9
+
+[metadata]
+license_file = LICENSE


cooperlees:
LGTM - Out of curiosity, why is including the license file important for you?

jdufresne:
> Out of curiosity, why is including the license file important for you?

For one, the license contains:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

So without this, flake8-bugbear appears to be violating its own license.

But more importantly, when I pull in 3rd party libraries into my projects, I like to be certain my project's license is compatible and I'm not violating the library's license. The more explicit the license is, the easier it is for me to determine this. Distributing the license with the wheel package, is very explicit. flake8-bugbear was just one example of packages included in my project not taking advantage of this.

PR #24, opened 2017-10-21
Add missing trove classifier for general Python 3 support

jdufresne:


cooperlees:
See no problem with this. Thanks!

PR #26, opened 2017-10-21
Remove Python 3.6-dev from Travis CI

jdufresne:
Python 3.6 is now stable and released. Can test against the released 3.6 instead of 3.6-dev. Testing both is redundant.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
     - 3.5
     - 3.6
-    - 3.6-dev
     - 3.7-dev
 install: pip install -e .
 script: python setup.py test


cooperlees:
@ambv - Does 3.6-dev still test alpha builds of the 3.6 branch? Does it have any value?

hugovk:
Yes, 3.6-dev is the 3.6 development branch:

https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions

There is still value testing against it:

https://snarky.ca/how-to-use-your-project-travis-to-help-test-python-itself/

So `3.6` is best, and `3.6` and `3.6-dev` is better.

ambv:
@hugovk has it right.

PR #27, opened 2017-10-21
Document flake8-bugbear as stable and ready for production use

jdufresne:


cooperlees:
See no problem with this and agree - we've had it in prod for a long time @ Facebook. Thanks!

PR #36, opened 2018-03-27
Fix pathlib link in readme

psykzz:
Previously without the schema, it would have an invalid link, creating a relative link within github.com
Adding the `https://` created the appropriate absolute link.


ambv:
Thanks! ✨ 🍰 ✨

PR #38, opened 2018-04-12
Add trove classifier 'Programming Language :: Python :: 3 :: Only'

jdufresne:
As the project only supports Python 3, should document this using ad trove classifier in setup.py.

For a full list of trove classifiers, see:

https://pypi.python.org/pypi?%3Aaction=list_classifiers


cooperlees:
I guess so. Thanks!

PR #45, opened 2018-06-06
Introduce B011

gaul:
Do not call `assert False` since `python -O` removes these calls.  Instead
callers should `raise AssertionError()`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -90,6 +90,9 @@ ahead of time.
 property access: ``x.attr = val``. There is no additional safety in
 using ``setattr`` if you know the attribute name ahead of time.
 
+**B011**: Do not call `assert False` since `python -O` removes these calls.
+Instead callers should `raise AssertionError()`.
+
 
 Python 3 compatibility warnings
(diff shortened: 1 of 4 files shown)


ambv:
That will raise false positives for users of pytest.  What situation made you want to create this warning?

gaul:
Sorry I do not follow how pytest triggers this?  I am concerned about `python -O` removing error handling from case statements like:

```python
if cond1:
    pass
elif cond2:
    pass
else:
    assert False, "unhandled"

**Comment by dmerejkowsky** [comment]
> Sorry I do not follow how pytest triggers this?

Sometimes you write code like this:

```
def test_something():
      if some_condition_no_met:
             assert False   # cause pytest to fail the test
```

You could argue people should use `pytest.fail()`, though

**Comment by gaul** [comment]
> Sometimes you write code like this:

Should authors instead write:

```python
def test_something():
      assert some_condition_no_met:
```

gaul:
> Please edit pull request so no longer contains reference to already taken B009 or close it if there is already a decision not to accept this feature.

Done.

PR #51, opened 2018-08-10
Consider using set / dict comprehensions

dmerejkowsky:
Stolen from pylint: https://github.com/PyCQA/pylint/blob/master/pylint/checkers/refactoring.py#L465

Some notes:

* There's already a pull request adding a 009 check, it's fine if you want me change it to 010

* I plan of "porting" more pylint warnings like this in the future (but don't worry, not everything at once!) Tell me if this OK with you.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -165,6 +165,9 @@ class BugBearVisitor(ast.NodeVisitor):
                 ):
                     self.errors.append(B004(node.lineno, node.col_offset))
 
+        if isinstance(node.func, ast.Name):
+            self.check_for_b009(node)
+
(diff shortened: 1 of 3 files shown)


dmerejkowsky:
Hum. It seems this is already done by https://pypi.org/project/flake8-comprehensions.

Feel free to close and sorry for the noise

cooperlees:
Many thanks for the idea and I understand why. Yes, flake8-comprehenions is awesome, I use it.

I think for bugbear, we actually want it to cause rather bad bugs in code to accept. Since we have another plugin that does this I will close it.

PR #52, opened 2018-09-05
Exclude immutable calls like tuple() and frozenset() from B008

jameshiew:
Uncommon to see this but these calls are OK as default argument nodes

### DIFF
--- a/tests/b006_b008.py
+++ b/tests/b006_b008.py
@@ -7,6 +7,10 @@ def this_is_okay(value=(1, 2, 3)):
     ...
 
 
+def and_this_also(value=tuple()):
+    pass
+
+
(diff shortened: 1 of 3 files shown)


tests/b006_b008.py
```diff
@@ -7,6 +7,10 @@ def this_is_okay(value=(1, 2, 3)):
     ...
 
 
+def and_this_also(value=tuple()):
```
ambv: Why would you ever want to say "tuple()" instead of just "()"?

cooperlees: I thought the same, but it is valid, all be it slower.

jameshiew: You wouldn't but I have seen code which does do this. It is a styling and performance issue rather than a B008.

PR #55, opened 2018-11-03
Introduce B009 and B010

joshkehn:
Using `getattr(x, 'foo')` and `setattr(x, 'foo', 'bar')` are no more
safe than normal property access and probably cause additional confusion
about safety when they are used because without a default `getattr`
raises the same `AttributeError` as dot access. There is also a
performance cost associated with doing attribute access indirectly.

I've added these as separate warnings since they opperate on two
different behaviors but they could probably be combined as a single
"fat" warning without trouble.

Example with `getattr`:

```python
>>> import os
>>> getattr(os, "foo")
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'foo'
```

Example with `setattr`:

```python
>>> os.foo = "bar"
>>> os.foo
'bar'
>>> setattr(os, "baz", "bar")
>>> os.baz
'bar'
```


README.rst
```diff
@@ -80,6 +80,16 @@ function will reuse the result of that definition-time call.  If this is
 intended, assign the function call to a module-level variable and use
 that variable as a default value.
 
+**B009**: Do not call ``getattr(x, 'attr')``, instead use normal
+property access: ``x.attr``. Missing a default to ``getattr`` will cause
+an ``AttributeError`` to be raised for non-existant properties. There is
```
carljm:
```suggestion
an ``AttributeError`` to be raised for non-existent properties. There is
```

PR #61, opened 2019-03-06
allow 'mcs' for metaclass classmethod first arg

davidism:
PyCharm's inspector wants the first arg of metaclass class methods to be called "mcs", it will complain if it's called "metacls" or any of the names Bugbear currently allows. I'd like to be able to satisfy both PyCharm and Bugbear. Since there dosen't appear to be a source of truth for this list of allowed names, I felt it was OK to add this one.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -233,6 +233,12 @@ MIT
 Change Log
 ----------
 
+19.3.0
+~~~~~~
+
+* For B902, the first argument for metaclass class methods can be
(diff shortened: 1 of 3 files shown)


davidism:
Turns out PyCharm's inspection allows choosing between "mcs" and "metacls", although it's not particularly discoverable.

cooperlees:
(I'll let this sit for a day or two incase anyone else has issues with this, if not, I'll merge and release to PyPI for you)

PR #75, opened 2019-06-19
Check shebang and file executability.

xuhdev:
- B012: There is a shebang in a Python file but the file is not executable. The developer may have forgotten to make it executable.
- B013: A Python file is executable but there is no shebang in it. This is quite common because very often developers make a file executable without noticing it (e.g., when using some unzip programs).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,6 +93,10 @@ using ``setattr`` if you know the attribute name ahead of time.
 **B011**: Do not call `assert False` since `python -O` removes these calls.
 Instead callers should `raise AssertionError()`.
 
+**B012**: Shebang is present but the file is not executable.
+
+**B013**: The file is executable but no shebang is present.
+
 
 Python 3 compatibility warnings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


cooperlees:
Hi,

Thanks for the patch. I come from an environment where we use the shebang to identify the file as Python 2 or Python 3, and even if that exists, we don't really need or want the file to be executable. This would be undesirable. I know we could just ignore this B012-13 in our flake8 file, but I don't feel like this should be in bugbear.

I'm not fully against this addon, but will need another bug bear maintainer to step up and say they want it for it to be merged. I've requested a couple to get their input.

Maybe a flake8-executable plugin could be made for this?

Thanks again,
Cooper

xuhdev:
@cooperlees After thinking this a bit, I agree that it might be a better idea to make a `flake8-executable` plugin. B012 does not make sense in your working environment; but for your working environment, we need a "strict mode" -- people normally shouldn't add executable permissions to Python files with or with shebangs, except those explicitly configured in `.flake8`.

If you agree, do you think flake8-executable is a good fit under PyCQA?

xuhdev:
I created [flake8-executable](https://github.com/xuhdev/flake8-executable). Feel free to check it out. Closing now.

Issue #67, opened 2019-04-23
B950 text is misleading


jcardali:
B950 checks for 1.1 * `max-line-length`, but the actual error text that is returned shows just `max-line-length`

https://github.com/PyCQA/flake8-bugbear/blob/master/bugbear.py#L48

ambv:
The point is for this to be a "highway speed limit" kind of thing. If you're overdoing it by a little, you *are* overdoing it but we are not nagging you. If you are overdoing it by a lot, we tell you the agreed max line length. If the speed limit is 65 mph, highway patrol is not going to stop you at 71 mph. But if you're going 80mph they will write a ticket for "80mph on 65mph", not "10 mph faster than reasonable".

If you want a hard limit, just use E501.

PR #76, opened 2019-06-24
Use adjusted_max_length in error message

mattbriancon:
Fixes https://github.com/PyCQA/flake8-bugbear/issues/67

Rounded the value to avoid an error message that included a float as the max line length.


ambv:
Well, this is not how I think about this. More explanation on the issue. Thanks for your contribution but we'll not be merging this.

PR #77, opened 2019-06-30
fix: Ignore yield if it's part of expr or __await__

### LINKED ISSUE #69: B901 should not apply to __await__ methods
Custom awaitable objects need to implement ``__await__`` generator methods with a ``return`` to provide values. B901 triggers on these for Python2 compatibility, but it is very unlikely that Python2 would use an ``__await__`` method. [The ``__await__`` method must be an iterator](https://docs.python.org/3/reference/datamodel.html#object.__await__), so using ``async def`` as suggested by B901 is not valid.

Marking every occurrence as ``# noqa: B901`` is very annoying in an async library. Disabling it entirely is suboptimal, because a generator-``return`` outside of ``__await__`` usually hints at an error indeed.

```python3
    def __await__(self) -> Generator[Any, Any, ST]:
        return (yield from self._await_message().__await__())
```

untitaker:
Fix #69


untitaker:
BTW, coming back to this project after a while it seems that it claims to be black-formatted, but the codebase doesn't seem to be formatted (black version 19.3 generated a large diff)

cooperlees:
It was formatted by myself with a very early version of black. Diffs are vary spontaneous with this project. I'll put another black PR up.

untitaker:
Could this one be reviewed?

cooperlees:
Looks good to me - Can you just rebase with my landing of 'blakening' more of the code and see that CI still passes and I'll merge.

CI will check for black.

untitaker:
Done!

cooperlees:
Thanks!

PR #78, opened 2019-07-01
Reformat with black + add to CI

cooperlees:
- Ran `find . -type f -name '*.py' | xargs black`
- CI will now error if not black formatted
- Add 3.8 testing


bugbear.py
```diff
@@ -130,7 +130,7 @@ def __getattr__(self, name):
     def visit(self, node):
         self.node_stack.append(node)
         self.node_window.append(node)
-        self.node_window = self.node_window[-self.NODE_WINDOW_SIZE:]
+        self.node_window = self.node_window[-self.NODE_WINDOW_SIZE :]
```
cooperlees: Why does black do this @ambv? Expected?

cooperlees:
Also, we need to get TravisCI working here again - I don't have access to work out why ...

cooperlees:
Waited long enough - Going to merge.

PR #93, opened 2019-10-14
Add explanation for not including lints in main lint tools

### LINKED ISSUE #92: Clarify why warnings "don't belong in pyflakes and pycodestyle"
Many of these seem like useful warnings, so I don't get why you say they don't belong in the main projects. Why not just make them disabled by default there?

cooperlees:
- Explain lints are opinionated and not standards
- Explain flake8 plugin design as the other reason

Fixes #92

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,6 +23,12 @@ pycodestyle::
     3. A fearsome imaginary creature, especially one evoked to frighten
        children.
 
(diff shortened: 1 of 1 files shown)


cooperlees:
@carljm + @ambv: Would love you to check my poor Englishes skillz. Ta.

PR #96, opened 2019-11-07
Removed a colon from the descriptive message in B008.

minusworld:
Hey! We are using bugbear wrapped inside a tool which parses flake8 messages for our code. There was a colon after B008, which we found when our parsing didn't look quite right. The colon was also inconsistent with the other flake8 messages... so this is a little PR that removes it.

Thanks! Great plugin!

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -497,7 +497,7 @@ B007 = Error(
     "If this is intended, start the name with an underscore."
 )
 B008 = Error(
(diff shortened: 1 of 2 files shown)


cooperlees:
Bah. It won’t let me merge. Can you just run black please. Then I’ll merge when Travis passes.

minusworld:
Fin!  Black was out of date.

Issue #99, opened 2019-12-31
Warn about continue/return/break in finally block


joaoe:
I reported this issue originally to pyflakes but they decided that is it not an error, and hence it should better be flagged as a warning by flake8-bugbear.

The issue:
```
try:
    raise ValueError(1)
finally:
  return # or break or continue
```
will silence the exception. Same with a `break` or `continue` statement in Python 3.8+, which has since been allowed in `finally` blocks. If there is a desire to silence the exception, they should explicitly be silenced in an `except` block. Putting a `if something: return` in a `finally` block is something I've done before, thinking that the exception would always be propagated, but it is actually not the case.

Another edge case
```
def a():
  try:
    return 1
  finally:
    return 2

a() = ?
```
Which value is returned ?

**References**
* original pyflakes issue about `return` in `finally` https://github.com/PyCQA/pyflakes/issues/404
* pyflakes adding support for `continue` in `finally` https://github.com/PyCQA/pyflakes/pull/476
* a PEP proposal to forbid return+continue in finally https://www.python.org/dev/peps/pep-0601/ 
* returns in finally are generally considered a bad practice. https://cwe.mitre.org/data/definitions/584.html

cooperlees:
This seems like a good candidate for bugbear. I’d love to not have this In any of my code.

I’d merge a PR for this, but would love @ambv to agree first. Please remember tests, or it will not be merged. Thanks.

joaoe:
I'm thinking about error code `B012`, enabled by default.
Any thoughts ?

PR #100, opened 2019-12-31
Warn about continue/return/break in finally block

joaoe:
Context:
Fixes : #99 

The new warning has code B012 and is enabled by default.

Bonus: refactored `check_for_b901()` to just use a recursive loop since I was initially trying to reuse that function, but turned out to be messy.


joaoe:
Ah, happens every time. And the code does not work for loops in the module
body.
For next year.

cooperlees:
Thanks. makes sense to me!

joaoe:
Thank you so much :)

PR #105, opened 2020-01-03
Add new checks for `except (...):` clauses

### LINKED ISSUE #98: Warn on useless entries in multi-type `except` clauses
When using a tuple of exception types in an `except` clause, it's easy for inexperienced Pythonistas to include elements that don't have any function at all.  For example:

```python
# TypeError is a subclass of Exception, so it doesn't add anything
try: ...
except (Exception, TypeError): ...

# Duplicate exception types are useless
try: ...
except (OSError, OSError): ...

# Literal single-element tuple is pointless complexity
try: ...
except (ValueError,): ...
```

It would be nice if `flake8-bugbear` could warn about these, or at least about the first!

Zac-HD:
When using a tuple of exception types in an `except` clause, it's easy for inexperienced Pythonistas to include elements that don't have any function at all.  For example:

```python
try: ...
except (Exception, TypeError): ...	 # TypeError is a subclass of Exception (builtins only)
except (MyError, MyError): ...  	 # Duplicate exception types are useless
except (MyError,): ...  			 # Should use bare name, not tuple
except (MyError, BaseException): ... # Even non-builtins subclass BaseException
```

This PR adds checks for all these cases, recommends the improved version (with `as`-clause if applicable), and thus closes #98.

CC @cooperlees

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -106,6 +106,13 @@ To silence an exception, do it explicitly in the `except` block. To properly use
 a `break`, `continue` or `return` refactor your code so these statements are not
 in the `finally` block.
 
+**B013**: A length-one tuple literal is redundant.  Write `except SomeError:`
+instead of `except (SomeError,):`.
+
+**B014**: Redundant exception types in `except (Exception, TypeError):`.
+Write `except Exception:`, which catches exactly the same exceptions.
+
+
 Python 3 compatibility warnings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -227,6 +234,12 @@ MIT
 
 Change Log
 ----------
+
+Future
+~~~~~~
+
+* For B001, also check for ``except ():``
+* Introduce B013 and B014 to check tuples in ``except (..., ):`` statements
 
 20.1.0
 ~~~~~~
--- a/bugbear.py
+++ b/bugbear.py
@@ -1,4 +1,5 @@
 import ast
+import builtins
 from collections import namedtuple
 from contextlib import suppress
 from functools import lru_cache, partial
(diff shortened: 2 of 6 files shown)


README.rst
```diff
@@ -228,6 +235,12 @@ MIT
 Change Log
 ----------
 
+Future
```
cooperlees: The Future!

bugbear.py
```diff
@@ -136,7 +137,45 @@ def visit(self, node):
...
+                #    (Exception, TypeError)  # builtins where one subclasses another
+                # but note that other cases are impractical to hande from the AST.
+                # We expect this is mostly useful for users who do not have the
+                # builtin exception hierarchy memorised, and include a 'shadowed'
+                # subtype without realising that it's redundant.
+                good = sorted(set(names), key=names.index)
+                if "BaseException" in good:
+                    good = ["BaseException"]
+                for name, other in itertools.permutations(tuple(good), 2):
+                    print(name, other)
```
Zac-HD: Oops, missed a print 😅

cooperlees: We all go this, lucky you noticed it.

I guess we could have CI look for weird output and assert if found.

Zac-HD: Or flake8-alfred and ban the print function?

cooperlees: Ban is such a strong word. Shouldn't you be in bed? :P Hope the fires are not near you mate. 🇦🇺❤️

Zac-HD: Well... I'm in Canberra, and no family in the way of the fires, which is a relief even with terrible air quality.  On the other hand the place my family camps every summer is in that black patch west of Cooma, the coastal area speaks for itself, and our Prime Minister is beyond hopeless :confused: 

image

On the upside, I've removed the stray `print`!

bugbear.py
```diff
@@ -539,6 +578,14 @@ def visit(self, node):
     "to be silenced. Exceptions should be silenced in except blocks. Control "
     "statements can be moved outside the finally block."
 )
+B013 = Error(
+    message="B013 A length-one tuple literal is redundant.  "
```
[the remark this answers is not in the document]
Zac-HD: Makes sense to me, but I've left it as "length-one tuple" to minimize the terminology that new programmers need to learn (right now to use the linter).

For example I have a `tox` config for formatters+linters that I encourage all my students to use when they're learning Python - in my experience automated feedback on what "good code" looks like can be really helpful, and the other bugbear messages are great at explaining *why* as well as *what*.

Zac-HD:
Turns out that this is useful for well-linted projects too:

> `hypothesis\internal\validation.py:85:5:` B014 Redundant exception types in `except (TypeError, OverflowError, ValueError, ArithmeticError):`.  Write `except (TypeError, ValueError, ArithmeticError):`, which catches exactly the same exceptions.

cooperlees:
Will merge and release tonight (US Pacific time)

PR #109, opened 2020-01-06
Fix error on attributes-of-attributes in `except (...):` clauses

Zac-HD:
CC @cooperlees and @AlexandreYang 

As well as fixing the bug, I've added some fuzz-tests that should catch any similar issues before release next time.  
(in this case they don't, meaning that (a) I need to improve Hypothesmith, and (b) there is no affected code in the standard library or any third-party packages in my dev environment.  Rare bug!)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setup(
     py_modules=["bugbear"],
     zip_safe=False,
     python_requires=">=3.6",
-    install_requires=["flake8 >= 3.0.0", "attrs"],
+    install_requires=["flake8 >= 3.0.0", "attrs>=19.2.0"],
     test_suite="tests.test_bugbear",
     classifiers=[
         "Development Status :: 5 - Production/Stable",
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
     - python: nightly
 
 install:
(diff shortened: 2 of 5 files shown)


Zac-HD:
> hypothesis 5.1.0 has requirement attrs>=19.2.0, but you'll have attrs 18.2.0 which is incompatible.

You can fix this for now by clearing the Travis cache, or fix it for good by pinning transitive dependencies with `pip-compile`.

cooperlees:
I’m always a fan of just moving forward so I feel we should just fix bugbear.py to work with >=19.2.0 if you’d like to use that module. I can’t image it being a huge PR.

Zac-HD:
Done - it already *works* with attrs>=19.2.0; not it *requires* that as a minimum version for easy compatibility with test dependencies.

setup.py
```diff
@@ -38,7 +38,7 @@
     py_modules=["bugbear"],
     zip_safe=False,
     python_requires=">=3.6",
-    install_requires=["flake8 >= 3.0.0", "attrs"],
+    install_requires=["flake8 >= 3.0.0", "attrs>=19.2.0"],
```
cooperlees: As this is not a hard requirement for flake8-bugbear, I might move this to .travis.yml if anyone opens an Issue about needing to run it with an older attrs.

PR #116, opened 2020-01-26
Silence B009/B010 for non-identifiers

### LINKED ISSUE #113: B009 false positive when attribute is not a keyword
When the attribute is not a valid Python token, use of `getattr` can be required; flake8-bugbear emits a B009.

Example:

    class DictProxy():

        def __init__(self, data):
            self._data = data

        def __getattr__(self, key):
            try:
                return self._data[key]
            except KeyError:
                raise AttributeError

    d = DictProxy({"foo": "bar", "123": "456"})
    print(d.foo)
    print(getattr(d, "123"))    # B009

The implied suggestion, `d.123`, would be a SyntaxError.

pkolbus:
If an object implements `__getattr__` and/or `__setattr__`, the second parameter to a `getattr()` or `setattr()` call might not be a valid Python identifier (e.g. `getattr(foo, "123abc")`). In this case bugbear should not emit B009/B010, since changing to `foo.123abc` would be a SyntaxError. Update the B009/B010 detection to check using a regex.

Fixes #113.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -5,6 +5,7 @@ from contextlib import suppress
 from functools import lru_cache, partial
 import itertools
 import logging
+import re
 
 import attr
 import pycodestyle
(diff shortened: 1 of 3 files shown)


pkolbus:
Thanks!

cooperlees:
Thanks for contributing! Appreciate eliminating all false positives!

PR #117, opened 2020-02-01
Don't emit B009/B010 for keywords

Zac-HD:
I wrote an argparse CLI with option named `except` - the code works fine with `getattr(ns, "except")`, but `ns.except` is a SyntaxError.  So this patch makes the linter keyword-aware :smile:

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -4,6 +4,7 @@ from collections import namedtuple
 from contextlib import suppress
 from functools import lru_cache, partial
 import itertools
+from keyword import iskeyword
 import logging
 import re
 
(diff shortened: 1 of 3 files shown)


Zac-HD:
:tada: 

New personal record: noticed problem to merged PR (on not-my-project) in less than 30 minutes!

cooperlees:
Pushed by a GitHub action to PyPI - https://github.com/PyCQA/flake8-bugbear/runs/420479491

Now if I could only automate generating README.rst changes + version change.

Even tho I'm in AU sweating I went with US date ...

Zac-HD:
Hey, Hypothesis automated those too - it's just a Python script away 😉

And commiserations, it's awfully hot here in Canberra 🔥

PR #129, opened 2020-06-28
Support exception aliases properly in B014

### LINKED ISSUE #110: B014 recommends `except ():`
In the following code:

```
    try:
        with open(fn) as f:
            return f.read().split('\n')
    except (OSError, IOError):
        pass
```

The new B014 warning shows this:

> B014 Redundant exception types in `except (OSError, IOError):`.  Write `except ():`, which catches exactly the same exceptions.

Doesn't seem to make sense to me. Is this a bug :bear:?

### LINKED ISSUE #126: Erroneous report message in B014
flake8 reports on this line
```python
    except (IOError, OSError, ImportError):
```
with
```
B014 Redundant exception types in `except (IOError, OSError, ImportError):`.  Write `except ImportError:`, which catches exactly the same exceptions.
```

I believe this message is incorrect, as is demonstrated by the following snippets

```python
try:
    raise OSError
except ImportError:
    print("gotcha")
# OSError is raised
```

```python
try:
    raise IOError
except ImportError:
    print("gotcha")
# IOError is raised
```

However
```python
try:
    raise OSError
except IOError:
    print("gotcha")
# ... "gotcha"
```

```python
try:
    raise IOError
except OSError:
    print("gotcha")
# ... "gotcha"
```

So  the original line is indeed redundant, but the report message is misleading

ichard26:
Resolves #110 
Resolves #126

```console
(env) richard-26@ubuntu-laptop:~/programming/flake8-bugbear$ cat test.py
try:
    with open("bugbear.py") as f:
        lines = f.readlines()
except (OSError, IOError):
    pass
(env) richard-26@ubuntu-laptop:~/programming/flake8-bugbear$ flake8 test.py
test.py:4:1: B014 Redundant exception types in `except (OSError, IOError):`.  Write `except OSError:`, which catches exactly the same exceptions.
```

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -174,6 +174,7 @@ class BugBearVisitor(ast.NodeVisitor):
                 #    (MyError, MyError)  # duplicate names
                 #    (MyError, BaseException)  # everything derives from the Base
                 #    (Exception, TypeError)  # builtins where one subclasses another
+                #    (IOError, OSError)  # IOError is an alias of OSError since Python3.3
                 # but note that other cases are impractical to hande from the AST.
                 # We expect this is mostly useful for users who do not have the
                 # builtin exception hierarchy memorised, and include a 'shadowed'
@@ -181,6 +182,16 @@ class BugBearVisitor(ast.NodeVisitor):
                 good = sorted(set(names), key=names.index)
                 if "BaseException" in good:
                     good = ["BaseException"]
+                # Find and remove aliases exceptions and only leave the primary alone
+                primaries = [
+                    primary
+                    for primary in B014.exception_aliases.keys()
+                    if primary in good
+                ]
+                for primary in primaries:
+                    aliases = B014.exception_aliases[primary]
+                    good = [e for e in good if e not in aliases]
+
                 for name, other in itertools.permutations(tuple(good), 2):
                     if issubclass(
                         getattr(builtins, name, type), getattr(builtins, other, ())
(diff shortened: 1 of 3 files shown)


bugbear.py
```diff
@@ -174,13 +174,24 @@ def visit_ExceptHandler(self, node):
...
                     good = ["BaseException"]
+                # Find and remove aliases exceptions and only leave the primary alone
+                primaries = [
+                    primary
+                    for primary in B014.exception_aliases.keys()
+                    if primary in good
+                ]
+                for primary in primaries:
+                    aliases = B014.exception_aliases[primary]
+                    good = [e for e in good if e not in aliases]
```
ichard26, line 193: Perhaps this is a bit over-engineered. I could just define an edge-case for OSError. 

Thoughts?

bugbear.py
```diff
@@ -174,13 +174,22 @@ def visit_ExceptHandler(self, node):
...
                 good = sorted(set(names), key=names.index)
                 if "BaseException" in good:
                     good = ["BaseException"]
+                # Find and remove aliases exceptions and only leave the primary alone
+                primaries = filter(
+                    lambda primary: primary in good, B014.exception_aliases.keys()
+                )
+                for primary in primaries:
+                    aliases = B014.exception_aliases[primary]
+                    good = list(filter(lambda e: e not in aliases, good))
```
ichard26, line 191: Perhaps this is a bit over-engineered. I could just define an edge-case for OSError.

Thoughts?

*Copy of original review because I modified this code afterwards to be more functional*

cooperlees: Na seems fine to me and makes sense.

PR #130, opened 2020-07-10
Add B015: pointless statement.

kalekseev:
For me the main reason to have such check is pytest style asserts, they are quite easy to miss in the middle of the test function

```python
def test_number of objects():
    ...setup code ...
    MyModel.objects.count() == 2   # should be `assert MyModel.objects.count() == 2`
    ...more asserts...
```

so tests are green and developer thinks he's asserted important results in the test.

Pylint has similar rule *pointless-statement* although it's more advanced but it's a pain to run pylint on tests and not all developers can tolerate false positives from it, so I think it would be a good addition to the flake8-bugbear.

Right now that rule checks only comparisons so maybe error should be: "pointless comparison" or we may try to make the rule more complete by including other cases, eg. uncalled functions:

```python
service = Service()
service.run  # should be service.run()
...
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,6 +117,9 @@ instead of ``except (SomeError,):``.
 **B014**: Redundant exception types in ``except (Exception, TypeError):``.
 Write ``except Exception:``, which catches exactly the same exceptions.
 
+**B015**: Pointless comparison. This comparison does nothing but
+wastes CPU instructions. Remove it.
+
 
 Python 3 compatibility warnings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/bugbear.py
+++ b/bugbear.py
@@ -1,16 +1,15 @@
 import ast
 import builtins
+import itertools
+import logging
+import re
 from collections import namedtuple
 from contextlib import suppress
 from functools import lru_cache, partial
-import itertools
 from keyword import iskeyword
-import logging
-import re
 
 import attr
 import pycodestyle
-
 
 __version__ = "20.1.4"
 
@@ -300,6 +299,9 @@ class BugBearVisitor(ast.NodeVisitor):
         self.check_for_b012(node)
         self.generic_visit(node)
 
+    def visit_Compare(self, node):
+        self.check_for_b015(node)
+
     def compose_call_path(self, node):
         if isinstance(node, ast.Attribute):
             yield from self.compose_call_path(node.value)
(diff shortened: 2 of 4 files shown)


cooperlees:
Also, lets make the name specific (i.e. pointless comparisons) since all it does is comparisons ...

kalekseev:
Done, though error looks a bit terse relative to others, not sure if we should add more details or is it too obvious to explain?

README.rst
```diff
@@ -117,6 +117,8 @@ instead of ``except (SomeError,):``.
 **B014**: Redundant exception types in ``except (Exception, TypeError):``.
 Write ``except Exception:``, which catches exactly the same exceptions.
 
+**B015**: Pointless comparison.
```
cooperlees, line 120:
```suggestion
**B015**: Pointless comparison. This comparison does nothing but wastes CPU instructions. Remove it.
```

Maybe just be a little more verbose here and give an action.

bugbear.py
```diff
@@ -658,6 +665,11 @@ def visit(self, node):
         "select.error",
     }
 }
+B015 = Error(
+    message=(
+        "B015 Pointless comparison. This comparison does nothing but wastes CPU instructions. Remove it."
```
cooperlees, line 670:
```suggestion
        "B015 Pointless comparison. This comparison does nothing but wastes "
        + "CPU instructions. Remove it."
```

README.rst
pquentin: In practice we're not worried about CPU instructions (especially in Python), but about missing asserts in pytest-style tests, so the wording is a bit strange here. How about "Did you forget an assert?" instead?

WhyNotHugo:
With the latest release, my CI system has picked up the latest version, and raised `B015` for a few of tests.

Just wanna say **thanks**: this change helped me find 8 missing `assert`s in unit tests that could've let bugs go unnoticed!

cooperlees:
Sir, Thankyou very much for sharing this. You often get complaints and not thanks. So it’s very nice to hear that everyone’s work here is helping people! Glad you have 8 more useful tests now!

pquentin:
Yes, we fixed a number of tests thanks to this warning. Thanks a lot! 💯

Zac-HD:
Just dropping by to say that this also caught a single missing `assert` in Hypothesis' tests!  Thanks to @kalekseev for this new check and @cooperlees for what I know is usually a thankless task maintaining things :smiling_face_with_three_hearts:

kalekseev:
Thank you guys for such a great feedback, I appreciate that!

PR #134, opened 2020-08-28
Remove check for # noqa comments

### LINKED ISSUE #44: Ignores all errors when using `noqa: X` to only ignore X
When using `# noqa: E501` all errors are excluded.

This is due to the code in https://github.com/PyCQA/flake8-bugbear/blob/6bf40eafe99bb569296c460b691a6c0bebd27ea5/bugbear.py#L39-L40.

`pycodestyle.noqa` searches for r'# no(?:qa|pep8)\b', which matches `# noqa: E501` also.

Without knowing the internals it seems like flake8 itself should include the errors itself - it has logic to ignore only specific errors (https://github.com/PyCQA/flake8/blob/013b925624bd5e4c026e83eb2c6e9f45cafa0dc0/src/flake8/defaults.py#L39-L51).

Example:
```python
s = 'some very, very, very, very, very, very, very, very, very, very, very, very, very, very, very long line'  # noqa: E123
```

flake8-bugbear should report error `t-noqa-colon.py:1:110: B950 line too long (109 > 80 characters)` for it, but is silent.

plinss:
Fixes #44

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -34,9 +34,6 @@ class BugBearChecker:
         visitor = self.visitor(filename=self.filename, lines=self.lines)
         visitor.visit(self.tree)
         for e in itertools.chain(visitor.errors, self.gen_line_based_checks()):
-            if pycodestyle.noqa(self.lines[e.lineno - 1]):
-                continue
-
             if self.should_warn(e.message[:4]):
                 yield self.adapt_error(e)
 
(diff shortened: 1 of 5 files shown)


cooperlees:
Here? https://github.com/PyCQA/flake8-bugbear/blob/master/.flake8#L5

plinss:
I don't think so, but it does look like I need to update some tests (was having trouble running the tests locally).

It looks like some tests are checking for the `# noqa` comments working when calling the BugBearChecker directly.

cooperlees (approved):
All makes sense to me with the explanation.

PR #136, opened 2020-09-04
Format with latest Black (20.08b1)

ichard26:


cooperlees:
Thought I'd done this. Cheers.

PR #141, opened 2020-10-17
Add check for raising a literal

gaul:
This is invalid and the user likely intended to return instead.
GitHub code search is not great but shows a few instances of this:

https://github.com/search?l=Python&p=1&q=%22raise+False%22&type=Code


cooperlees (approved):
I like it, thanks!

If you see this before I merge - Can you add the change to “Change Log” in README.rst

gaul:
> If you see this before I merge - Can you add the change to “Change Log” in README.rst

Done.

cooperlees:
Thanks!

PR #142, opened 2020-10-17
Add missing call to generic_visit

gaul:
Matches other call sites.


cooperlees (approved):
Thanks - I wonder if we can test for this to ensure others don’t miss this ...

PR #144, opened 2020-11-13
Exclude types.MappingProxyType() from B008

hukkinj1:
Exclude the read-only types.MappingProxyType() from B008 similar to tuple() and frozenset().


cooperlees (approved):
Thanks. This makes sense to me since we're >= 3.6. I'll make sure CI passes and then merge. We're due for a release this month. I'll try spend some time this weekend.

cooperlees:
If you want to help speed me up - Please add the change to README.rst please.

hukkinj1:
:+1: Updated the README

Issue #154, opened 2021-01-27
Exclude float("inf") from B008


erichang-bcad:
Currently, using positive/negative infinity as default values for function arguments will trigger B008, as it thinks we're calling the float() function.  Unless there are other ways to specify infinities as literals in Python, I think this behavior should be fixed.  Thanks.

cooperlees:
Yeah I'd accept a fix with that. Should be able to say if function name == float ignore. Please also unittest and comment why this was done for future code visitors. Thanks.

cooperlees:
If there is int() or other type equivalents, we might as well cover them too.

erichang-bcad:
I can submit a PR, but I have a few questions about implementation:

1. do you want to allow calls for ``int()``?  In Python, only positive/negative infinities and NaN don't have literals, so you don't need to allow ``int()`` if you want to always force users to write out string literals.

2. do you want a relaxed or strict check?  If you want to be strict, you can only allow ``float()`` if argument is a string representing infinities or NaN.  If you want to be relaxed, you can skip checking the argument of ``float()``.

3. If you want to allow ``int()``, do you want a relaxed or strict check?  A strict check will make sure the argument is a string literal (and optionally the ``base`` parameter).

cooperlees:
> I can submit a PR, but I have a few questions about implementation:
> 
> 1. do you want to allow calls for `int()`?  In Python, only positive/negative infinities and NaN don't have literals, so you don't need to allow `int()` if you want to always force users to write out string literals.

Rigthtio. Let's skip int. I wasn't sure if it made sense.

> 2. do you want a relaxed or strict check?  If you want to be strict, you can only allow `float()` if argument is a string representing infinities or NaN.  If you want to be relaxed, you can skip checking the argument of `float()`.

Hmm - I think it's better to start strict for this edge case. See if anyone else shares a preference. Can always go looser if it makes sense.

> 3. If you want to allow `int()`, do you want a relaxed or strict check?  A strict check will make sure the argument is a string literal (and optionally the `base` parameter).

Ignore since we can skip int in this PR.

Thanks!

ichard26:
Uhhhhhhhhhhhhhhhhhhhh, this is the *Close Issues That Should Be But Aren't* police here,

PR GH-155 has been merged so yeahhh... hi @cooperlees 

(and no I *don't* regret the overcasualiness I'm using)

cooperlees:
Thanks @ichard26

PR #155, opened 2021-01-28
make sure float infinity/NaN does not trigger B008

erichang-bcad:
This PR addresses issue #154, and make sure float infinity/NaN does not trigger a B008 error.  Some unit tests are added.

I tested this code on Python 3.9, I don't know how to test this on Python 3.7 or below (note that this is needed because the ast
module behavior for constants changed), do you have regressions that check across versions automatically?

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -3,6 +3,7 @@ import builtins
 import itertools
 import logging
 import re
+import sys
 from collections import namedtuple
 from contextlib import suppress
 from functools import lru_cache, partial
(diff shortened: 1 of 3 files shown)


cooperlees:
We do have CI. I don't know why it triggered. I clicked re-run and it seems to be running now. Lets see how things go in the older versions of python. Thanks!

cooperlees (approved):
Aweosme, thanks! All LGTM with good test coverage. Will leave for a day or two for others to weigh in if they wish.

PR #158, opened 2021-02-16
Handle positional-only args in class methods (fixes #157)

### LINKED ISSUE #157: B902 false positive: positional-only arguments in class methods
The following code triggers B902 on the `bar` method:

```python
class Foo:
    def bar(self, a, /):
        pass
```

bryanforbes:


cooperlees (approved):
LGTM! Thanks.

Awesome tests for the simple change.

(Will leave for 24 hours to see if any other maintainer weighs in)

PR #161, opened 2021-03-05
Hotfix issue 153

### LINKED ISSUE #153: crash on tuple expansion in try/except block
Running flake8 against the following file produces no report as an exception is raised in bugbear
```python
# dummy.py
grey_list = (ValueError,)
black_list = (TypeError,)

try:
    int("1e3")
except (*grey_list, *black_list):
    print("error caught")
```

crashes with
```
...
  File "/Users/clm/.pyenv/versions/3.8.6/envs/yt_dev/lib/python3.8/site-packages/bugbear.py", line 126, in _to_name_str
    assert isinstance(node, ast.Attribute)
AssertionError
```

For information
```
$ flake8 --version
3.8.4 (flake8-bugbear: 20.11.1, mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 3.8.6 on Darwin
```

neutrinoceros:
This is an attempt to fix #153
This is my very first experience with `ast` and I'll admit I don't really know what I'm doing. I added a test based on the issue's description, checked that it failed with the current state of the code base as reported, and applied a minimal patch to make it pass without breaking any other test.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -124,8 +124,10 @@ def _to_name_str(node):
     # "pkg.mod.error", handling any depth of attribute accesses.
     if isinstance(node, ast.Name):
         return node.id
-    assert isinstance(node, ast.Attribute)
-    return _to_name_str(node.value) + "." + node.attr
+    try:
+        return _to_name_str(node.value) + "." + node.attr
+    except AttributeError:
+        return _to_name_str(node.value)
 
 
 def _typesafe_issubclass(cls, class_or_tuple):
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Thanks! LGTM, This handles the condition and stops us crashing. I can't see a better way to do this. If anyone else can we can forward fix!

neutrinoceros:
Thank you for merging !
When do you think I can realistically expect this fix to be released (no rush, just curious) ?

cooperlees:
Yeah, I have it all automated apart from changes in README.rst - So releasing is easy here. Just need to not let it sit so long so we ship more with less change.

PR #163, opened 2021-03-30
Add B017 - detection for an evil form of assertRaises

cricalix:
```with assertRaises(Exception):``` is basically a "catch 'em all" assert that casts far too broad of a net when it comes to detecting failures in code being tested. Assertions should be testing specific failure cases, not "did Python throw /any/ type of error?", and so the context manager form, or the assertRaisesRegex form are far better to use.

With the implementation of this code, it is possible that someone would write ```with self.assertRaises(Exception) as ex:``` and then never check ```ex```, but linters should then warn about the unused context variable.

W503 had to be ignored because ```test_selfclean_bugbear()``` was upset by the warnings being raised for the syntax in use, despite that syntax being the syntax created by the recommended formatter, black.

This kind of lint check can also be found in openstack/hacking as H202, but implementing that module for flake8 requires pulling in a lot of extras, and it's very opinionated about versions of the extras to import.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -12,7 +12,7 @@ from keyword import iskeyword
 import attr
 import pycodestyle
 
-__version__ = "21.3.2"
+__version__ = "21.3.3"
 
 LOG = logging.getLogger("flake8.bugbear")
 
@@ -317,6 +317,10 @@ class BugBearVisitor(ast.NodeVisitor):
         self.check_for_b016(node)
         self.generic_visit(node)
 
+    def visit_With(self, node):
+        self.check_for_b017(node)
+        self.generic_visit(node)
+
     def compose_call_path(self, node):
         if isinstance(node, ast.Attribute):
             yield from self.compose_call_path(node.value)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: bugbear.py [hunk 1]
```diff
@@ -423,6 +427,26 @@ def check_for_b016(self, node):
...
+        Exception, which happens to be the vast majority of Python internal
+        errors, including the ones raised when a non-existing method/function
+        is called, or a function is called with an invalid dictionary key
+        lookup.
+        """
+        item = node.items[0]
+        item_context = item.context_expr
+        if (
+            hasattr(item_context.func, "attr")
+            and item_context.func.attr == "assertRaises"  # noqa W503
```
cooperlees, line 443:
```suggestion
            and (item_context.func.attr == "assertRaises")  # noqa W503
```
Does that remove W503 firing?

tests/b017.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Foobar(unittest.TestCase):
+    def evil_raises(self) -> None:
+        with self.assertRaises(Exception):
+            raise Exception("Evil I say!")
+
+    def context_manager_raises(self) -> None:
+        with self.assertRaises(Exception) as ex:
+            raise Exception("Context manager is good")
+        self.assertEqual("Context manager is good", str(ex.exception))
```
cooperlees, line 16:
```suggestion
            raise Exception("Context manager is good")
            self.assertEqual("Context manager is good", str(ex.exception))
```
I feel if this was ran, this would throw a ValueError ? Not important, but yeah.

cooperlees (changes requested):
Looks good to me. Can you just:
- Remove version change (who knows when I will release, but will be soon)
- See if you can get rid of W503, if not no big deal
- Fix the context manager raises whitespace

Then also
- Add to README your explanation like other checks
  - Add this to "Change Log" in README too please

bugbear.py
cricalix: Nope. W503 continues to fire. As I understand it, the current format I've got (that black applied) is PEP8 compliant, and W503 is wrong, for some value of wrong?

tests/b017.py
cricalix: Nope, ex is left in scope when the context manager exits. Been writing my tests like this for years, and Python has never thrown a ValueError.

cooperlees: Hmm, that seems to go against the whole context vibe. But sure. I just try not to do it to avoid bugs as I've seen variations of this bite.

cooperlees (approved):
All looks good to me and ready to merge.

PR #167, opened 2021-04-02
Revamp B017 conditionals and expand coverage of problems

cricalix:
This diff re-addresses #164 by completely rewriting the logic for finding the assertRaises code in the AST.

It also adds support for finding cases of Exception when Exception is in the tuple of things that are being checked by the assertRaises.

Conditional logic was based on output from ast.dump.

For the bare exception case:
```
    context_expr=Call(
        func=Attribute(
            value=Name(id='self', ctx=Load()),
            attr='assertRaises',
            ctx=Load()),
        args=[
            Name(id='Exception', ctx=Load())],
        keywords=[]))
```

For the tuple of builtin exceptions case:
```
    context_expr=Call(
        func=Attribute(
            value=Name(id='self', ctx=Load()),
            attr='assertRaises',
            ctx=Load()),
        args=[
            Tuple(
                elts=[
                    Name(id='Exception', ctx=Load()),
                    Name(id='ValueError', ctx=Load())],
                ctx=Load())],
        keywords=[]))
```

For the tuple of builtin + module exceptions case:
```
    context_expr=Call(
        func=Attribute(
            value=Name(id='self', ctx=Load()),
            attr='assertRaises',
            ctx=Load()),
        args=[
            Tuple(
                elts=[
                    Name(id='Exception', ctx=Load()),
                    Attribute(
                        value=Name(id='asyncio', ctx=Load()),
                        attr='CancelledError',
                        ctx=Load())],
                ctx=Load())],
        keywords=[]))
```

Negative test cases were also checked using ast.dump, and also include other ``with`` syntax.

A sample ast tree for the negative case of a context manager assertRaises:
```
    context_expr=Call(
        func=Attribute(
            value=Name(id='self', ctx=Load()),
            attr='assertRaises',
            ctx=Load()),
        args=[
            Name(id='Exception', ctx=Load())],
        keywords=[]),
    optional_vars=Name(id='ex', ctx=Store()))
```

Examples in the tests/b017.py file are legit broken (or working) code.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -4,6 +4,7 @@ import itertools
 import logging
 import re
 import sys
+import typing as t
 from collections import namedtuple
 from contextlib import suppress
 from functools import lru_cache, partial
@@ -13,7 +14,7 @@ import attr
 
 import pycodestyle
 
-__version__ = "21.4.3"
+__version__ = "21.4.4"
 
 LOG = logging.getLogger("flake8.bugbear")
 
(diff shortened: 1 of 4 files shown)


bugbear.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import re
 import sys
+import typing as t
```
cricalix, line 7: Wasn't sure whether to use ``from __future__ import annotations`` and use ``nodes: list`` at line 442, or to use this form, so picked this form.

cooperlees: We can't support that due to 3.6 support I think.
```suggestion
from typing import List
```

I really dislike `import typing as t`

`def hard_to_read_typing(x: t.List[t.Dict[t.Union[str, bool]]]) -> t.Dict[t.List[t.Any]]:`

is terrible to read imo compared to:

`def readable_typing(x: List[Dict[Union[str, bool]]]) -> Dict[List[Any]]:`

bugbear.py
```diff
@@ -437,18 +438,31 @@ def check_for_b017(self, node):
...
-            and hasattr(item_context.func, "attr")  # noqa W503
-            and item_context.func.attr == "assertRaises"  # noqa W503
-            and len(item_context.args) == 1  # noqa W503
-            and isinstance(item_context.args[0], ast.Name)  # noqa W503
-            and item_context.args[0].id == "Exception"  # noqa W503
-            and not item.optional_vars  # noqa W503
-        ):
-            self.errors.append(B017(node.lineno, node.col_offset))
+
+        def exception_finder(nodes: t.List) -> bool:
```
cooperlees, line 442:
```suggestion
        def exception_finder(nodes: List) -> bool:
```

cooperlees (approved):
More testing and handling sounds good. Will leave to see if anyone else swings by to review.

bugbear.py
cricalix: I had thought that there might be a version issue. At least we can use typing.

cooperlees:
Ok, sorry I forgot about this. Having my second look, should we leave the old unit tests to ensure we support them or not? The rewrite confuses me a little bit.

cooperlees:
@cricalix - What to merge this? I'm a little confused so either message me and we can talk it through or I think I'm going to close the PR and leave it as is since we've had no complaints. Some of this rewrite I don't understand.

cricalix:
I’m at sea/off the grid ish until September. I’ll look then.

cooperlees:
Reopen / re-submit if you want to do this :)

PR #173, opened 2021-06-19
B008: Whitelist additional immutable function calls

### LINKED ISSUE #172: B008: Whitelist operator.{attrgetter,itemgetter,methodcaller} default args
Calling those functions in default args is perfectly fine and *not* calling them would most likely be a bug.

ThiefMaster:
closes #172

I did not whitelist unqualified `compile` since pretty much everyone uses `re.compile` and `compile` is also a built-it with a completely different meaning.


cooperlees (approved):
Thanks. I agree with the compile choice. This makes less noise, so I support it.

Will leave it for a few days to see if anyone else comments before I merge.

PR #181, opened 2021-08-22
Add `B904` - Use `raise ... from err` in `except` clauses

### LINKED ISSUE #180: Proposed check: explicit `raise` in `except` or `finally` should use `from ...`
Thanks to [PEP 3134](https://www.python.org/dev/peps/pep-3134/), Python 3 supports [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining).  For example,

```python
try:
    assert False
except AssertionError as err:
    raise RuntimeError("An assertion failed")  # Should use `from err` or `from None`
```
```python-traceback
  File "t.py", line 2, in <module>
    assert False
AssertionError

During handling of the above exception, another exception occurred:

  File "t.py", line 4, in <module>
    raise RuntimeError("An assertion failed")
RuntimeError: An assertion failed
```

The line `During handling of the above exception, another exception occurred:` is not strictly speaking *wrong*, but for explicit `raise` statements we can do better:

- If we want to *replace* the earlier exception, we should `raise ... from None` to omit it from the traceback
- If we want to *add more information*, we should `raise ... from err` - the message would then be `The above exception was the direct cause of the following exception:`

*This issue was prompted by reading https://github.com/numpy/numpy/issues/15986 and wishing for a better tool than `grep` to check my own code.*

Zac-HD:
Closes #180, by adding a new check ~~B018~~ B904 to recommend [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining).  For example,

```python
try:
    assert False
except AssertionError:
    raise RuntimeError  # B904: Within an except clause, use `raise ... from err` ...
```

This turned out to be remarkably easy to add to the existing code 😁

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,7 +2,7 @@
 # Keep in sync with setup.cfg which is used for source packages.
 
 [flake8]
-ignore = E203, E302, E501, E999
+ignore = E203, E302, E501, E999, W503
 max-line-length = 88
 max-complexity = 12
 select = B,C,E,F,W,B9
--- a/README.rst
+++ b/README.rst
@@ -123,9 +123,9 @@ waste CPU instructions. Either prepend ``assert`` or remove it.
 **B016**: Cannot raise a literal. Did you intend to return it or raise
 an Exception?
 
-**B017**: ``self.assertRaises(Exception):`` should be considered evil. It can lead 
-to your test passing even if the code being tested is never executed due to a typo. 
-Either assert for a more specific exception (builtin or custom), use 
+**B017**: ``self.assertRaises(Exception):`` should be considered evil. It can lead
+to your test passing even if the code being tested is never executed due to a typo.
+Either assert for a more specific exception (builtin or custom), use
 ``assertRaisesRegex``, or use the context manager form of assertRaises
 (``with self.assertRaises(Exception) as ex:``) with an assertion against the
 data available in ``ex``.
@@ -220,7 +220,7 @@ Change Log
 21.4.3
 ~~~~~~
 
-* Verify the element in item_context.args is of type ast.Name for b017 
+* Verify the element in item_context.args is of type ast.Name for b017
 
 21.4.2
 ~~~~~~
@@ -231,7 +231,7 @@ Change Log
 ~~~~~~
 
 * Add B017: check for gotta-catch-em-all assertRaises(Exception)
-  
+
 21.3.2
 ~~~~~~
 
--- a/README.rst
+++ b/README.rst
@@ -130,6 +130,11 @@ Either assert for a more specific exception (builtin or custom), use
 (``with self.assertRaises(Exception) as ex:``) with an assertion against the
 data available in ``ex``.
 
+**B018**: Within an ``except`` clause, raise exceptions with ``raise ... from err``
+or ``raise ... from None`` to distinguish them from errors in exception handling.
+See [the exception chaining tutorial](https://docs.python.org/3/tutorial/errors.html#exception-chaining)
+for details.
+
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
@@ -216,6 +221,11 @@ MIT
 
 Change Log
 ----------
+
+Future
+~~~~~~
+
+* Add B018: check for ``raise`` without ``from`` in an ``except`` clause
 
 21.4.3
 ~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -130,11 +130,6 @@ Either assert for a more specific exception (builtin or custom), use
 (``with self.assertRaises(Exception) as ex:``) with an assertion against the
 data available in ``ex``.
 
-**B018**: Within an ``except`` clause, raise exceptions with ``raise ... from err``
-or ``raise ... from None`` to distinguish them from errors in exception handling.
-See [the exception chaining tutorial](https://docs.python.org/3/tutorial/errors.html#exception-chaining)
-for details.
-
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
@@ -161,6 +156,11 @@ data classes that only set attributes in an ``__init__`` method, and do
 nothing else. If the attributes should be mutable, define the attributes
 in ``__slots__`` to save per-instance memory and to prevent accidentally
 creating additional attributes on instances.
+
+**B904**: Within an ``except`` clause, raise exceptions with ``raise ... from err``
+or ``raise ... from None`` to distinguish them from errors in exception handling.
+See [the exception chaining tutorial](https://docs.python.org/3/tutorial/errors.html#exception-chaining)
+for details.
 
 **B950**: Line too long. This is a pragmatic equivalent of
 ``pycodestyle``'s E501: it considers "max-line-length" but only triggers
@@ -225,7 +225,7 @@ Change Log
 Future
 ~~~~~~
 
-* Add B018: check for ``raise`` without ``from`` in an ``except`` clause
+* Add B904: check for ``raise`` without ``from`` in an ``except`` clause
 
 21.4.3
 ~~~~~~
--- a/bugbear.py
+++ b/bugbear.py
@@ -292,7 +292,7 @@ class BugBearVisitor(ast.NodeVisitor):
 
     def visit_Raise(self, node):
         self.check_for_b016(node)
-        self.check_for_b018(node)
+        self.check_for_b904(node)
         self.generic_visit(node)
 
     def visit_With(self, node):
@@ -427,7 +427,7 @@ class BugBearVisitor(ast.NodeVisitor):
         ):
             self.errors.append(B017(node.lineno, node.col_offset))
 
-    def check_for_b018(self, node):
+    def check_for_b904(self, node):
         """Checks `raise` without `from` inside an `except` clause.
 
         In these cases, you should use explicit exception chaining from the
@@ -440,7 +440,7 @@ class BugBearVisitor(ast.NodeVisitor):
             and not (isinstance(node.exc, ast.Name) and node.exc.id.islower())
             and any(isinstance(n, ast.ExceptHandler) for n in self.node_stack)
         ):
-            self.errors.append(B018(node.lineno, node.col_offset))
+            self.errors.append(B904(node.lineno, node.col_offset))
 
     def walk_function_body(self, node):
         def _loop(parent, node):
@@ -762,13 +762,6 @@ B017 = Error(
         "context manager form of assertRaises."
     )
 )
-B018 = Error(
-    message=(
-        "B018 Within an `except` clause, raise exceptions with `raise ... from err` "
-        "or `raise ... from None` to distinguish them from errors in exception handling.  "
-        "See https://docs.python.org/3/tutorial/errors.html#exception-chaining for details."
-    )
-)
 
 # Warnings disabled by default.
 B901 = Error(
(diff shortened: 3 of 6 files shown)


README.rst
```diff
@@ -123,13 +123,18 @@ waste CPU instructions. Either prepend ``assert`` or remove it.
...
+Either assert for a more specific exception (builtin or custom), use
 ``assertRaisesRegex``, or use the context manager form of assertRaises
 (``with self.assertRaises(Exception) as ex:``) with an assertion against the
 data available in ``ex``.
 
+**B018**: Within an ``except`` clause, raise exceptions with ``raise ... from err``
+or ``raise ... from None`` to distinguish them from errors in exception handling.
+See [the exception chaining tutorial](https://docs.python.org/3/tutorial/errors.html#exception-chaining)
+for details.
+
```
cooperlees, line 137: Do we need to care to make this only care if we're >= 3.9?
```suggestion
**B018**: Within an ``except`` clause, raise exceptions with ``raise ... from err``
or ``raise ... from None`` to distinguish them from errors in exception handling (i.e. prefer exception chaining).
See [the exception chaining tutorial](https://docs.python.org/3/tutorial/errors.html#exception-chaining)
for details.

```

cooperlees (changes requested):
Looks good - Thanks!

My only questions are (And this is open to others):
- Should this be optional?
- Should this be only on >= 3.9 (Is that when this feature was introduced?)

Also, what triggers W503? I'm guessing some code you didn't add ...
- Can we # nova: W503 them so we can look to clean them up?

README.rst
```diff
@@ -123,13 +123,18 @@ waste CPU instructions. Either prepend ``assert`` or remove it.
 **B016**: Cannot raise a literal. Did you intend to return it or raise
 an Exception?
 
-**B017**: ``self.assertRaises(Exception):`` should be considered evil. It can lead 
-to your test passing even if the code being tested is never executed due to a typo. 
-Either assert for a more specific exception (builtin or custom), use 
+**B017**: ``self.assertRaises(Exception):`` should be considered evil. It can lead
+to your test passing even if the code being tested is never executed due to a typo.
+Either assert for a more specific exception (builtin or custom), use
```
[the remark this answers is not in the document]
Zac-HD: Yeah, stripping trailing whitespace.  I can delete that commit if you prefer.

README.rst
Zac-HD: Exception chaining was added in Python 3.0!  It's just not in widespread use yet, due to the syntatic incompatibility with Python 2 😅

.flake8
```diff
@@ -2,7 +2,7 @@
 # Keep in sync with setup.cfg which is used for source packages.
 
 [flake8]
-ignore = E203, E302, E501, E999
+ignore = E203, E302, E501, E999, W503
```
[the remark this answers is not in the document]
Zac-HD: Following [`black`'s recommendation](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html?highlight=W503#flake8) - W503 is actually disabled by default, but enabled for bugbear by the `W` in `select = B,C,E,F,W,B9`.  I updated the config because I *also* added the first case of this pattern, [in `check_for_b018()` here](https://github.com/PyCQA/flake8-bugbear/pull/181/files#diff-21e991c5e10f0aca96647b43b4d0968e75c52cc2bceaacf0794f50ded872e49bR461-R465).

README.rst
cooperlees: O - I went to docs here: https://docs.python.org/3.9/tutorial/errors.html#exception-chaining
- Set it to 3.8 and this section disappears ...

cooperlees (approved):

cooperlees:
Will give it till the morning pacific - If no one says anything against this I'll merge.

README.rst
Zac-HD: Wow, I guess it's not been well documented... [the reference docs](https://docs.python.org/3.8/library/exceptions.html) still don't have a heading for it, though `ctrl-f chain` does find some mentions.

Zac-HD:
I've added one cheeky exception for `raise some_error`, because both Hypothesis and Trio use this pattern when constructing an exception object with a specific traceback - and it seems unlikely to silence the warning on code which *is* of concern.

.flake8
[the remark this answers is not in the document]
hukkin: Maybe use `extend-ignore` instead of `ignore` here to not override the defaults?

`W503` for instance is off by default unless defaults are overridden by `ignore` configuration.

hukkin:
> Should this be optional?

I do think that the `from e` syntax should _in theory_ always be used when raising within an exception handler. _In practice_ however, I find myself too lazy to write out the extra characters, as the benefit is very little, especially if the re-raised exception will be caught later on and never bubbles up to a traceback. Sometimes the added characters make the line long enough that it needs to be wrapped in more than one line making the expression slightly less readable.

Just some reasoning why making this optional should perhaps be considered. Personally don't feel strongly either way.

Zac-HD:
Fair points @hukkin - you can always disable it in global config though, and IMO if something is off by default it might as well not exist for most users 😕

hukkin:
Yeah I agree that off-by-default warnings are not much use.

Thinking a bit more about this from a library developer perspective, there's two types of errors that will be raised:
- Errors that are raised when a downstream developer (consumer of the library) makes incorrect use of the API --> we want the best traceback possible
- Errors that are handled internally within the library and should never be seen by downstream developers --> we don't care about the traceback

What would be great is if this warning could only be applied to the first error type, but that isn't possible really. So I feel the question is really which one is more important: having an error for the first error type, or not needlessly annoying developers raising errors of the second error type.

Application developers also may find this annoying as they should generally shut down gracefully and never let an error bubble all the way up.

Then there's the argument that all software has bugs, so any error may bubble up even though not intended to do so...

Generally I think I'm starting to lean more towards "this is probably more annoying than useful".

Zac-HD:
[Here's the diff in Hypothesis](https://github.com/HypothesisWorks/hypothesis/compare/bf2c4be6f86ccb88a496373b24e92ad8e270bc01...72a27341064df83cd380a3c5b6821b9628f2e55c), which has plenty of both examples:

| Patched     | `from err` | `from None` |
| :---:       |      :---: |       :---: |
| user-facing |          2 |           6 |
| internal    |          4 |           1 |

Personally I think that even the internal changes are net-positive, in that the code now expresses that these exceptions were raised *because of* rather than *while handling* the previous error, and that the user-facing improvements are enough to be worth some inconvenience anyway.

It's easy to turn off the check if you don't like it, but exception chaining is *amazingly* under-used simply because very few people know it exists and it wasn't viable for libraries until we stopped supporting Python 2 - even just saying "hey, this exists" is useful!

cooperlees:
> I think the best way to understand this is to try out the three different ways of exception chaining:
> 
> #### Disable chaining
> ```python
> >>> try:
> ...     raise Exception("one")
> ... except Exception:
> ...     raise Exception("two") from None
> ... 
> Traceback (most recent call last):
>   File "<stdin>", line 4, in <module>
> Exception: two
> ```
> 
> #### Implicit chaining
> ```python
> >>> try:
> ...     raise Exception("one")
> ... except Exception:
> ...     raise Exception("two")
> ... 
> Traceback (most recent call last):
>   File "<stdin>", line 2, in <module>
> Exception: one
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "<stdin>", line 4, in <module>
> Exception: two
> ```
> 
> #### Explicit chaining
> ```python
> >>> try:
> ...     raise Exception("one")
> ... except Exception as e:
> ...     raise Exception("two") from e
> ... 
> Traceback (most recent call last):
>   File "<stdin>", line 2, in <module>
> Exception: one
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "<stdin>", line 4, in <module>
> Exception: two
> ```
> 
> Why I think the feature isn't that widely used (in addition to Python 2 compat) is that the default (implicit chaining) is pretty good: it doesn't hide information. I fear that rather than receiving thanks, we may receive complaints for an overly opinionated check with this one.

I kind of also like the default behavior and am leaning towards hukkin's points now. So we have people each way here. What do we want to do? I think there is noise here. We could make it optional for people who want it @Zac-HD ?

Zac-HD:
I'd still prefer it on-by-default, but you're the maintainer - if you want it optional, I'll make it optional 🙂

cooperlees:
I would love a second maintainer (all thought most are not active these days) thoughts if any swing by ....

Zac-HD:
@asottile @The-Compiler @graingert - do you think we should recommend `raise ... from err` by default, or only if specifically enabled?

asottile:
some questions:

- without `from` you still get exception chaining, but with `from` it changes the message slightly right?
- how does this interact with `from_traceback`?
- will individuals respond to the error incorrectly and slap `from None` on things?  (many times the correct response is `raise e` => `raise`)

(I actually don't think I have the commit bit here, but stop by from time to time so feel free to discount my questions / opinions)

The-Compiler:
To get an idea of the impact, I ran this over both qutebrowser (~160 instances) and pytest (~300 instances). Both probably wouldn't profit too much from the new warning: qutebrowser because it's an application so it hardly matters; pytest due to how it's in charge of how it prints its own tracebacks anyways.

I guess flake8-bugbear users appreciate a more nitpicky linter compared to "plain" flake8 users, but on the other side that's often a reason people [don't like pylint](https://github.com/PyCQA/pylint/issues/746), because despite all configurability, at some point it drives people away if a linter produces more noise than signal by default.

I suppose it boils down to discoverability vs. a low-noise default configuration. For some projects I suspect this will be rather useless as others said, but of course bugbear doesn't know which ones those would be.

FWIW I've also rarely seen `raise ... from ...` in the wild, so even if useful, it's not exactly a best practice (yet?) I suppose.

I'm avoiding a concrete answer to the question because I really don't know. I'd end up disabling the check and not mind much, but others might feel differently.

Zac-HD:
OK; that's enough to convince me that disabled-by-default is sensible - and change made, it's now `B904`.

> * without `from` you still get exception chaining, but with `from` it changes the message slightly right?

Yep, you get exception changing regardless, `from` just improves the message (or `from None` hides earlier tracebacks).

> * how does this interact with `from_traceback`?

If you mean `dis.Bytecode.from_traceback()` (or `BaseException.with_traceback()`), it's handled the same way regardless of `from`.  [The docs](https://docs.python.org/3/library/exceptions.html) aren't great, but introspection uses the `__context__` attribute which isn't affected - `from` just sets the `__cause__`.

> * will individuals respond to the error incorrectly and slap `from None` on things?  (many times the correct response is `raise e` => `raise`)

Quite possibly.  `raise e` wouldn't trigger this check though; since I allow raising an all-lowercase name.  This does suggest another new lint rule though...

cooperlees (approved):
Thanks. I think this will be safer from noise. If people come round wanting this on by default we can look at changing.

Will try update CHANGES etc. + release this week.

yuwash:
Sorry I’m really not sure about. With the controversy already shown above, I think it should be off by default. It only “improves” the phrasing of the error message a bit although chaining itself happens by default. It seems rather hacky to try to change the wording by introducing such a linter rule rather than making the change to python itself. The [referenced part of the docs](https://docs.python.org/3/tutorial/errors.html#exception-chaining) notably doesn’t say one should always use `from`. It only says “useful when you are transforming exceptions” whereas I’m not completely sure what counts as “transforming”.

Issue #171, opened 2021-06-02
AttributeError: 'Call' object has no attribute 'value'


jaap3:
I added flake8-bugbear to a project that could use some cleanup, but upon running flake8 it crashed with a `AttributeError: 'Call' object has no attribute 'value'`.

I boiled down the offending code to this reproducer:

```python
from django.contrib.auth import get_user_model


def get_user_by_email(email):
    try:
        return get_user_model().get(email=email)
    except (
        get_user_model().DoesNotExist,
        get_user_model().MultipleObjectsReturned
    ):
        pass
```

This is with the following version of flake8/plugins/python:

```
flake8 --version
3.9.2 (flake8-bugbear: 21.4.3, mccabe: 0.6.1, pycodestyle: 2.7.0, pyflakes:
2.3.1) CPython 3.7.10 on Linux
```

szabgab:
I got the same error when I tried to run flake8 with bugbear installed on the source code of https://github.com/ansible/ansible

jaap3:
Here's the shortest reproducer I can come up with that generates the same stacktrace :

```
foo = lambda: IOError

try:
    ...
except (foo(),):
    ...
```

jaap3:
This is now fixed

szabgab:
Thank you. I just checked and the most recent version (21.9.2) does not raise an exception on the source code of ansible.

cooperlees:
Awesome. Thanks for report/confirming the fix worked!

PR #187, opened 2021-09-28
Do not crash on call in except statement

jaap3:
This fixes the crash reported in issue #171. Not sure if this is the correct way to fix this, but at least it doesn't crash.

### DIFF
--- a/tests/test_bugbear.py
+++ b/tests/test_bugbear.py
@@ -336,6 +336,14 @@ class TestFuzz(unittest.TestCase):
         )
         BugBearVisitor(filename="<string>", lines=[]).visit(syntax_tree)
 
+    def test_does_not_crash_on_call_in_except_statement(self):
+        # akin to test_does_not_crash_on_tuple_expansion_in_except_statement
+        # see https://github.com/PyCQA/flake8-bugbear/issues/171
+        syntax_tree = ast.parse(
+            "foo = lambda: IOError\ntry:\n    ...\nexcept (foo(),):\n    ...\n"
+        )
+        BugBearVisitor(filename="<string>", lines=[]).visit(syntax_tree)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/bugbear.py
+++ b/bugbear.py
@@ -125,6 +125,8 @@ def _to_name_str(node):
     # "pkg.mod.error", handling any depth of attribute accesses.
     if isinstance(node, ast.Name):
         return node.id
+    if isinstance(node, ast.Call):
+        return _to_name_str(node.func)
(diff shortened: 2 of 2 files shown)


cooperlees (approved):
Looks right to me. If CI passes I'll merge and look at getting a version out.

Many thanks for the Test driven fix! Welcome to flake8-bugbear family of contributors.

cooperlees (changes requested):
Bah - Sorry to be a pain, but our good friend black is uphappy with the formatting. Can you just fix that.

jaap3:
Super weird, I already ran black on it. Just ran it again and I get no changes:

```
black .
All done! ✨ 🍰 ✨
25 files left unchanged.
```

```
$ black --version
black, version 21.9b0
```

cooperlees:
> Super weird, I already ran black on it. Just ran it again and I get no changes:
> 
> ```
> $ black .
> All done! ✨ 🍰 ✨
> 25 files left unchanged.
> ```
> 
> ```
> $ black --version
> black, version 21.9b0
> ```

We're running and testing out the "experimental-string-processing" feature.

So you'll need: `black --experimental-string-processing .`

jaap3:
Ah, just noticed that as well. Should be good now.

tests/test_bugbear.py
```diff
@@ -336,6 +336,14 @@ def test_does_not_crash_on_tuple_expansion_in_except_statement(self):
         )
         BugBearVisitor(filename="<string>", lines=[]).visit(syntax_tree)
 
+    def test_does_not_crash_on_call_in_except_statement(self):
+        # akin to test_does_not_crash_on_tuple_expansion_in_except_statement
+        # see https://github.com/PyCQA/flake8-bugbear/issues/171
+        syntax_tree = ast.parse(
+            "foo = lambda: IOError\ntry:\n    ...\nexcept (foo(),):\n    ...\n"
```
cooperlees, line 343: I too, don't like this autoformat ...

cc: @ichard26 @JelleZijlstra @zsol - I think this will get complaints ... thoughts?

cooperlees (approved):

tests/test_bugbear.py
ichard26: Aren't user splits preserved unless they violate the length limit according to the ESP rules?

JelleZijlstra: Maybe the issue is that the string put together isn't long enough to hit the line length limit, so we put it all together.

jaap3: I think this is similar to https://github.com/psf/black/issues/1540

PR #188, opened 2021-09-28
Remove executable flags

jaap3:
Just noticed that this test file had mode 0755 while the rest of the files have 0644, I don't see a reason why this file in particular should be executable.


cooperlees (approved):
Thanks. Someone must have done by accident.

PR #189, opened 2021-09-30
Update setup.py

Roger-cmd:


cooperlees (approved):
Thanks for the attention to detail.
- I wonder if we should automate this / fail CI when this does not mate up to current year ...

PR #191, opened 2021-10-05
B904: ensure the raise is in the same context with the except

### LINKED ISSUE #190: B904 `from None` check gets confused when class defined in `except` has a `raise`
In Werkzeug, we want to handle the case where Python was not compiled with SSL, so we `try import ssl` and define a placeholder class in `except` if the import fails. The class will `raise` for any attribute access. B904 gets confused and thinks it should be written `raise ... from None` even though that doesn't make sense here.

```python
try:
    import ssl
except ImportError:

    class _SslDummy:
        def __getattr__(self, name: str) -> t.Any:
            raise RuntimeError("SSL support unavailable")  # noqa: B904

    ssl = _SslDummy()
```

Note that I had to add `# noqa: B904` to the `raise` line.

isidentical:
Resolves #190. Introduces a simple version of context management, though more detailed analysis might need something a little bit specialized in the future (e.g [stuff like this](https://github.com/isidentical/refactor/blob/1178e9faee21227b5f73f8815724cf9bae048200/refactor/context.py#L165-L177)).

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -510,7 +510,7 @@ class BugBearVisitor(ast.NodeVisitor):
                 break
 
     def check_for_b902(self, node):
-        if not len(self.contexts) >= 2 or not isinstance(
+        if len(self.contexts) < 2 or not isinstance(
             self.contexts[-2].node, ast.ClassDef
         ):
             return
--- a/bugbear.py
+++ b/bugbear.py
@@ -181,7 +181,7 @@ class BugBearVisitor(ast.NodeVisitor):
     node_window = attr.ib(default=attr.Factory(list))
     errors = attr.ib(default=attr.Factory(list))
     futures = attr.ib(default=attr.Factory(set))
-    context = attr.ib(default=attr.Factory(list))
+    contexts = attr.ib(default=attr.Factory(list))
 
     NODE_WINDOW_SIZE = 4
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -16,6 +16,19 @@ import pycodestyle
 __version__ = "21.9.2"
 
 LOG = logging.getLogger("flake8.bugbear")
+CONTEXTFUL_NODES = (
+    ast.Module,
+    ast.ClassDef,
+    ast.AsyncFunctionDef,
+    ast.FunctionDef,
+    ast.Lambda,
+    ast.ListComp,
+    ast.SetComp,
+    ast.DictComp,
+    ast.GeneratorExp,
+)
+
+Context = namedtuple("Context", ["node", "stack"])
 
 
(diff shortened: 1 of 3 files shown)


cooperlees:
@Zac-HD - Thoughts?

@isidentical - Sorry - we run black formatting with an experimental feature:
- To fix: `black --check --experimental-string-processing .`

isidentical:
@cooperlees ah, my bad. By the way, is there a particular reason why this project doesn't use `pre-commit` (or some other script) to do these stuff? If there is none, I can create a patch about it.

Zac-HD:
The concept sounds good to me, though the diff looks like it's patching B902 instead of B904?

(this also seems pretty rare to me, but as a general rule "if someone provides a patch, it matters enough to merge")

isidentical:
No actually not, I do not need to change anything on the B904's implementation, since it just checks if there are any except handlers within the current `node_stack`. What I did was to dynamically determine the current node stack from the context (through a property). I also needed the alter the B902 to work with the new implementation but no changes on it's behavior.

davidism:
Very occasionally, a more nested pattern comes up, such as trying a backport import as well. I don't think I actually have any code that does this, but would this change also avoid the lint message in this case?

```python
try:
    from preferred_library import Thing
except ImportError:
    try:
        from fallback_library import Thing
    except ImportError:
        class Thing:
            def __getattr__(self, name):
                raise AttributeError
```

cooperlees:
This is only failing on black. I'll fix before merging. If you get time, please reformat. Otherwise I'll take care of it.

cooperlees:
@isidentical - Sorry it's taken me so long but now this with a rebase to master is unhappy. Do you have any cycles to clean this up? Was looking to release today. I'll see if I can understand too now or I might release without this and we just cut another release when you get time.

bugbear.py
```diff
@@ -168,6 +181,7 @@ class BugBearVisitor(ast.NodeVisitor):
     node_window = attr.ib(default=attr.Factory(list))
     errors = attr.ib(default=attr.Factory(list))
     futures = attr.ib(default=attr.Factory(set))
+    context = attr.ib(default=attr.Factory(list))
```
cooperlees, line 184: Is this all that is needed? How did this ever work if so :|
```suggestion
    contexts = attr.ib(default=attr.Factory(list))
```

cooperlees (changes requested):

cooperlees (approved):
Ok - That makes sense to me. Will give @isidentical a while to respond. No idea how that `self.contexts` got changed to `self.context` ... Unless it was a refactor / push error.

Issue #200, opened 2021-10-28
B018: Find more constants w/o assign


kasium:
As discussed in #195, I took a look what B018 could potential find, and basically that are all Literals 😄 : https://greentreesnakes.readthedocs.io/en/latest/nodes.html#literals. Also, I guess that check could be extended to also work on module level

@cooperlees : I'm not sure how to do that test wise. Should I just list an example of each in a class and a function? Like

````python
def foo2():
    """my docstring"""
    a = 2
    "str"  # Str
    1j  # Number (complex)
    1  # Number (int)
    1.0  # Number (float)
    b"foo"  # Binary
    f"{int}"  # JoinedStr
    ...  # Ellipsis
    True  # NameConstant (True)
    False  # NameConstant (False)
    None  # NameConstant (None)
```

PR #201, opened 2021-10-29
B018: Find more constants w/o assign

kasium:
This check now also finds:
* numbers (ints, float, complex)
* lists
* sets
* dicts
* binary strings
* joined strings (f-strings)
* name constants (True, False, None)

This is part 1 of #200 . The second part will enable the check for modules


cooperlees (approved):
SGTM - This makes b018 much more complete.  Thanks for the efforts. /me wonders how many weird typos across the world this will pickup.

I'll look to release over the weekend / Monday (west coast USA) time ...

asottile:
I think this still misses a few things, such as this subtle bug:

```python
def f():
    f"string"
```

kasium:
@asottile thanks for your feedback. If I'm not wrong, JoinedStr is catched:
https://github.com/PyCQA/flake8-bugbear/blob/master/bugbear.py#L589

And a f-string w/o parameter is catched in pyflakes I guess

asottile:
it's only caught at body positions 1+

kasium:
Ah, yes. You are right.

However, I'm not sure if this should be covered here. It's not recommended to use `"` as the docstring and to find out which quotes where used, I guess tokens are needed. Therefore, I see this more in https://pypi.org/project/flake8-docstrings/.

@cooperlees what do you think?

asottile:
no all I'm suggesting is to extend the check to 0+ instead of 1+ and add an exception for Str at 0

kasium:
Sorry, that was clear 😄. Let me open a PR. That's a great idea!

PR #202, opened 2021-11-15
Improve B018 further

kasium:
Catch also useless expressions on the first node in a body if the node is not a string (docstring).
Thanks to @asottile for the idea.

Part of #201


cooperlees (approved):
Seems to make sense to me. Thanks all. Will wait and allow some others to disagree and then try and get a release out this week.

kasium:
Cool, thanks!

Issue #203, opened 2021-11-19
Option to specify immutable classes


finswimmer:
Hello,

my feature request was already mentioned [here](https://github.com/PyCQA/flake8-bugbear/issues/62#issuecomment-653668505) by @DylanYoung. Because the issue was about another topic, I thought it would be a good idea to put it into a new ticket.

I'd like to have an option to specify immutable classes. My concrete use case is `fastapi`. When you define your endpoints you have something like `Depends()` or `Query()` in the function calls, which results in a `B008` error message by `flake8-bugbear`.

Of course one can mark those lines with a `# noqa: B008`, but depending on how large the application is, this is not quite.

fin swimmer

PR #204, opened 2021-11-22
Add Option to extend the list of immutable calls

finswimmer:
This PR adds in option to extend the list of immutable calls. Calls to these method as argument defaults no longer raises a `B008` warning.

This is useful when using library like `fastapi` which comes with their own set of immutable calls.

Closes: https://github.com/PyCQA/flake8-bugbear/issues/203


cooperlees (approved):
Nice addition as people are always going to create more immutable calls etc. etc. so like the flexibility here.

finswimmer:
Cool, thanks :+1: Any plans when a new release will arrive?

cooperlees:
Released now.

PR #206, opened 2021-11-24
B014: catch binascii.Error and ValueError redundancy + cleanup

### LINKED ISSUE #178: B014: binascii.Error and ValueError redundancy not caught
I found an issue with flake8 not raising any flags for the following code:
```
try:
   ...
except (ValueError, binascii.Error):
    pass
```

whereas the error ```B014 Redundant exception types in `except (ValueError, UnicodeDecodeError):`.  Write `except ValueError:`, which catches exactly the same exceptions.``` is raised for the code below:
```
try:
   ...
except (ValueError, UnicodeDecodeError):
    pass
```

I find it strange as both `UnicodeDecodeError` and `binascii.Error` both inherit from `ValueError`

Is it a known issue?

ichard26:
I am honestly a bit embarrassed by my old code. Let's fix it up with better comments and clearer names :slightly_smiling_face: 

Fixes #178

--- 

FWIW Cooper, I wouldn't mind joining this project as a maintainer. I can't promise much as I'm already busy enough with the projects I'm leading right now, but you _are_ running an one-man show here :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -224,6 +224,12 @@ MIT
 Change Log
 ----------
 
+Unreleased
+~~~~~~~~~~
+
+* Update B014: ``binascii.Error`` is now treated as a subclass of ``ValueError``
+  (#206)
+
 21.9.2
 ~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


cooperlees (approved):
Thanks for the cleanup and explanations everywhere making it all easy to understand!

PR #209, opened 2021-11-29
Remove detection of strings in B018

### LINKED ISSUE #208: B018 wrongly detects inline variable or attribute docstrings
Having an inline attribute doc string or a module variable docstring, it is wrongly marked as B018 (sample from sphinx doc):

```
module_level_variable2 = 98765
"""int: Module level variable documented inline.

The docstring may span multiple lines. The type may optionally be specified
on the first line, separated by a colon.
"""

```

kasium:
Stops false positive for #208 issue - We need to evaluate if we can be smarter and re-add this detection in future releases.

### DIFF
--- a/tests/b018_classes.py
+++ b/tests/b018_classes.py
@@ -1,6 +1,6 @@
 """
 Should emit:
-B018 - on lines 15-26, 30, 32
+B018 - on lines 16-26, 30, 33
 """
 
 
@@ -30,3 +30,4 @@ class Foo3:
     123
     a = 2
     "str"
+    1
--- a/bugbear.py
+++ b/bugbear.py
@@ -639,7 +639,6 @@ class BugBearVisitor(ast.NodeVisitor):
             if isinstance(
                 subnode.value,
                 (
-                    ast.Str,
                     ast.Num,
                     ast.Bytes,
                     ast.NameConstant,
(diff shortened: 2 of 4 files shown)


tests/b018_classes.py
```diff
@@ -30,3 +30,4 @@ class Foo3:
     123
     a = 2
     "str"
+    1
```
cooperlees, line 33: Did we error at all without these ints? Why did you add them? Just trying to understand.

cooperlees (approved):
Thanks for the workaround. I'd like to release possibly today a less noisy version for people until we workout if we can re-enable strings and avoid this edge case.

I feel we could add an example string to the unittests to ensure we're actually now ignoring it. And then use it for the upcoming true fix (if we can).

tests/b018_classes.py
kasium: As you can see in the tests line 32 is ignored and to test if something is found, I added line 33

kasium: So this is already a working test

cooperlees: Yeah, this tests we're ignoring strs as a whole, I was more stating while we have examples that are causing the noise lets add those into this file too. But it can be next PR as well.

cooperlees (approved):

tests/b018_classes.py
kasium: Ah, you mean the docstring. Valid point. I'll take a look what we can improve

PR #213, opened 2021-12-31
Allow the same special cases for B950 as E501 (#176)

### LINKED ISSUE #176: B950: Propose exemption for long URLs
I like the idea of using B950 over E501, but I do not want to be warned anytime I include a long URL in a comment. Breaking URLs over multiple lines is not acceptable as I want them to be clickable.

pycodestyle addresses this by making an exemption to the line length error for comments that contain no whitespace. See [line 304](https://github.com/PyCQA/pycodestyle/blob/aa3417b6a51f5912e32d9c8c879e1b9dd660d5f8/pycodestyle.py#L304) (currently) of pycodestyle.py:

```
# Special case for long URLs in multi-line docstrings or
# comments, but still report the error when the 72 first chars
# are whitespaces.
chunks = line.split()
if ((len(chunks) == 1 and multiline) or
    (len(chunks) == 2 and chunks[0] == '#')) and \
        len(line) - len(chunks[-1]) < max_line_length - 7:
    return
```

I propose adding a similar exemption to flake8-bugbear. Of course, I could add a `# noqa` comment after every URL comment, but I much prefer pycodestyle's solution.

percurnicus:
Fixes #176 

Allows for the same special cases as E501 for B950

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -12,7 +12,7 @@ from keyword import iskeyword
 import attr
 import pycodestyle
 
-__version__ = "21.11.29"
+__version__ = "21.12.0"
 
 LOG = logging.getLogger("flake8.bugbear")
 CONTEXTFUL_NODES = (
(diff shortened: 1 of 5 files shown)


bugbear.py
```diff
@@ -12,7 +12,7 @@
 import attr
 import pycodestyle
 
-__version__ = "21.11.29"
+__version__ = "21.12.0"
```
cooperlees, line 15: Lets not do this here and I'll do it on a diff when I release (as it's going to be 22.*)

bugbear.py
```diff
@@ -67,8 +67,31 @@ def gen_line_based_checks(self):
...
             if length > 1.1 * self.max_line_length:
+                # Special case long URLS and paths to follow pycodestyle.
+                # Would use the `pycodestyle.maximum_line_length` directly but
+                # need to supply it arguments that are not available so chose
+                # to replicate instead.
+                chunks = line.split()
+
+                is_line_comment_url_path = len(chunks) == 2 and chunks[0] == "#"
+
+                just_long_url_path = len(chunks) == 1
```
cooperlees, line 84: So if I read this right, really this is just a long no white space string right? So it could not be a URL or file path? I guess generally we'll have spaces I guess is the rationale here right?

Is this the same as how pycodestyle does it? if so, (Sorry - being lazy and not checking), then I guess it's proven to be safe enough.

cooperlees (changes requested):
Thanks for this. Tests look sane to test the code. Would like the version change removed (as that version is never going to exist) and just answer my question about the code logic.

bugbear.py
percurnicus: Yeah, a long string with no white space, which is mostly just URLs and paths. A path that has a space in it will still fail but would also fail E501. Rational is it is more readable to not split up URLs and paths in comments and code.

It's the same logic as pycodestyle, modified to be a bit more readable to me. Linking [here](https://github.com/PyCQA/pycodestyle/blob/main/pycodestyle.py#L267-L296) for convenience.

bugbear.py
percurnicus: addressed

cooperlees (approved):

PR #214, opened 2022-01-07
Build universal Python 3 wheels

johnthagen:
Pure Python 3 packages need not include any configuration for `bdist_wheel`. Currently the wheels on PyPI could look to someone new as if they only support Python 3.6-3.8

- https://pypi.org/project/flake8-bugbear/21.11.29/#files

But CI is in fact testing through 3.10:

https://github.com/PyCQA/flake8-bugbear/blob/2095bdd2293b5567ea1dbeb485cab7da13a2e2fe/.github/workflows/ci.yml#L11

Supported Python versions are enforced through:

https://github.com/PyCQA/flake8-bugbear/blob/2095bdd2293b5567ea1dbeb485cab7da13a2e2fe/setup.py#L42

This change will build proper `*-py3-none-any.whl ` wheels.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[bdist_wheel]
-python-tag = py36.py37.py38
-
 [flake8]
 # Keep in sync with .flake8. This copy here is needed for source packages
 # to be able to pass tests without failing selfclean check.


cooperlees (approved):
Agree with this and I’ve been meaning to do it. Also need to move to the new wheel building tool (who’s name I forget - someone moved black). Thanks for this. I’ll merge and try get a release out soon.

johnthagen:
> Also need to move to the new wheel building tool (who’s name I forget - someone moved black). 

@cooperlees You are probably referring to the new [`build`](https://pypi.org/project/build/) tool.

An example of how to use it (very easy):

- https://github.com/johnthagen/python-blueprint/blob/b816b51bb55a37402d462bc76089af3b5bc5b883/tox.ini#L93

cooperlees:
That’s the one. Will accept PR if anyone has spare cycles!

johnthagen:
Here is the link to what needs to be added to `pyproject.toml` to support this:

- https://setuptools.pypa.io/en/latest/build_meta.html#how-to-use-it

A blog post about the topic in general that is a good read: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

Issue #217, opened 2022-01-12
Fail lru_cache / cache applied to methods


adamchainz:
Applying either decorator to methods caches instances longer than they would otherwise live, causing memory usage and maybe other bugs. Detailed explanation:  https://youtu.be/sVjtp6tGo0g

I propose a rule to catch this potential error

adamchainz:
Brilliant to see this merged. Thanks @sco1 and @cooperlees !

Issue #215, opened 2022-01-07
B008 false positive in FastAPI / Typer


johnthagen:
In FastAPI it is very common to use class objects as special, succicient indicators to the framework.

- https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#use-query-as-the-default-value

```py
@app.get("/items/")
async def read_items(q: Optional[str] = Query(None, max_length=50)):
```

Typer exposes the same, convenient convenient user experience:

- https://typer.tiangolo.com/tutorial/arguments/optional/#an-alternative-cli-argument-declaration

```py
def main(name: str = typer.Argument(...)):
    typer.echo(f"Hello {name}")


if __name__ == "__main__":
    typer.run(main)
```

`flake8-bugbear` will throw `B008` errors at this usage. It would be nice if `flake8-bugbear` whitelisted these or allowed the user to whitelist certain types.

johnthagen:
I now see that `extend-immutable-calls` exists. I think the docs could use an example for how this could be used (if possible?) on third-party types.

I would be happy to submit a PR if someone can give me a lead on how this configuration could be used with FastAPI or Typer types.

johnthagen:
```ini
[flake8]
...
extend-immutable-calls =
    Argument
```

This worked for the `typer` example. I will submit a PR if a maintainer can confirm that this is intended/correct usage for the problem in this issue.

npip99:
If anyone is using `pyproject.toml`,

```toml
[tool.ruff.lint.flake8-bugbear]
extend-immutable-calls = [
    # Whitelisted default arguments
    "fastapi.Header",
    "fastapi.Depends",
]
```

Issue #212, opened 2021-12-06
extend-immutable-calls: how to configure?


DavidVujic:
I haven't been able to make the [recently released feature](https://github.com/PyCQA/flake8-bugbear/pull/204) `extend-immutable-calls` to work. 

Have I configured it wrong? 

I have verified that `flake8` recognizes the config, by setting the line-length to 10 and verified flake8 warnings about line length.

```bash
flake8 --version

4.0.1 (flake8-bugbear: 21.11.29, mccabe: 0.6.1, pycodestyle: 2.8.0, pyflakes: 2.4.0) CPython 3.10.0 on
Darwin
```

`example.py`
```python
import fastapi


def example_fn(data: str = fastapi.Header(None)):
    pass
```

`setup.cfg`
```
[flake8]
max-line-length = 120
extend-immutable-calls = ["fastapi.Header"]
```

```bash
flake8 --show-source

./example.py:4:28: B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.
def example_fn(data: str = fastapi.Header(None)):
                           ^
```

aberres:
Same problem here - I tried to silence `pathlib.Path`.

sco1:
You want:

```
[flake8]
max-line-length = 120
extend-immutable-calls = fastapi.Header
```

cooperlees:
Do we feel we should add anything more to the README here? I'd accept a PR if so. Otherwise do we close?

sco1:
Even though it's just `ConfigParser` doing it's thing, it wouldn't hurt to add an example to the README (see also: #215).

I'm happy to include in #218 if you feel it's appropriate.

aberres:
Ah, great thanks.

And the matching happens by name, right?

So in the following case I would need `extend-immutable-calls = Path` to silence the warning.

```
from pathlib import Path

def foo(bar=Path()):
   ...
```

aberres:
And as we are here already: Would it make sense to add a hint to this parameter to the error message of B008?

DavidVujic:
Great, thank you for clarifying! 

I had thought that the values should be defined as a (Python) array. Now, when typing something like this (for more than one fastAPI thing to allow), everything works as expected:
```
extend-immutable-calls = fastapi.Header, fastapi.Depends
```

cooperlees:
[...]

PR #218, opened 2022-01-13
Add B019 check to find cache decorators on class methods

sco1:
As described in #217, using `functools.cache`/`functools.lru_cache` on class methods can potentially cause their instances to live longer than expected.

I took a stab at adding a rule to catch this pattern but I have a few questions:

1. `flake8-bugbear`'s error messages generally try to propose a solution to make the linting error go away (besides ignoring the line), but I'm not sure about the verbiage to include. For reference, the starting point is the following:

> Use of `functools.lru_cache` or `functools.cache` on class methods can lead to memory leaks. The cache may retain instance references, preventing garbage collection.

2. Should this be restricted to caches that are explicitly unbounded, rather than any use of a built-in cache with a class method? i.e. this would only raise for `functools.cache` and `functools.lru_cache(maxsize=None)`. This seems to more specifically address the issue though some still may not want to potentially keep these instances alive, bounded cache or not.

Related reading:
* https://bugs.python.org/issue19859 
* https://bugs.python.org/issue44310
* https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls

Resolves: #217
Resolves: #215
Resolves: #212
Resolves: #221


sco1:
For 2, I saw that a [FAQ entry](https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls) was added a few months ago (via: [bpo-44310](https://bugs.python.org/issue44310)) which might help craft some (brief) guidance?

asottile:
that whole `should_warn` should probably get removed, flake8 has had `extend-ignore` for a while now making `ignore =` mostly obsolete.  there's also `extend-select` in the latest version which makes the `select` stuff not needed either

sco1:
At first glance I think dropping it is pretty easy, mostly just removing a conditional layer here:

https://github.com/PyCQA/flake8-bugbear/blob/1c47a162d4b2c41ada9164a12b536d6f49833d7e/bugbear.py#L60-L62

Which still passes the test suite.

Given that this method is an intentional design decision:

> I think this behavior is surprising so Bugbear's opinionated warnings require explicit selection.

If its removal is desired, what would be the best way to update the language in the README, and what version of `flake8` should it target? I believe the [configuration options section](https://github.com/PyCQA/flake8-bugbear#how-to-enable-opinionated-warnings) was last touched before `extend-ignore` was introduced to `flake8` (3.6?) and `extend-select` landed in 4.0.

cooperlees:
So I think we should look at moving to all the latest select hotness, my only worry is will we break existing configuration for users of flake8-bugbear - Is there anything we can do to support both for now and I can do a release advertising the deprecation of things we need to deprecate?

sco1:
The filter method can be left in & just made aware of `--extend-select` (if available), which should address #221. Even with it remaining for a deprecation period I think the configuration verbiage should still change to recommend `--extend-select` and `--extend-ignore`.

Depending on the answer to question 3 in the OP, the method's `lru_cache` can just be `noqa`ed until the method is deprecated; I don't anticipate it being likely to cause any issues in this application.

cooperlees:
> The filter method can be left in & just made aware of `--extend-select` (if available), which should address #221. Even with it remaining for a deprecation period I think the configuration verbiage should still change to recommend `--extend-select` and `--extend-ignore`.
> 
> Depending on the answer to question 3 in the OP, the method's `lru_cache` can just be `noqa`ed until the method is deprecated; I don't anticipate it being likely to cause any issues in this application.

Yeah I'm happy to start recommending the new and lets state we're deprecating the old in README + Change log (even tho its same file) as I think some tooling parse out he change log + I paste into the GitHub Release page.

sco1:
I've updated the README to (hopefully!) capture all the possible configuration caveats.

I've also included a few minor fixes for other issues:
  * Example `extend-immutable-calls` configuration (#212, #215)
  * Make the warning filter aware of `extend-select` (#221)
  
I've edited the OP down to the remaining questions I have on this draft implementation.

README.rst
```diff
@@ -189,41 +192,56 @@ on the first line and urls or paths that are on their own line::
...
-disabled for B950 to be hit.
-
-If you'd like all optional warnings to be enabled for you (future proof
-your config!), say ``B9`` instead of ``B901``. You will need Flake8 3.2+
-for this feature.
-
-Note that ``pycodestyle`` also has a bunch of warnings that are disabled
-by default.  Those get enabled as soon as there is an ``ignore =`` line
-in your configuration.  I think this behavior is surprising so Bugbear's
+To enable these checks, specify a ``--extend-select`` command-line option or
```
cooperlees, line 195:
```suggestion
To enable bugbear (B) checks, specify a ``--extend-select`` command-line option or
```

README.rst
```diff
@@ -189,41 +192,56 @@ on the first line and urls or paths that are on their own line::
...
+``extend-select=`` option in your config file (requires flake8 4.0+)::
+
+  [flake8]
+  max-line-length = 80
+  max-complexity = 12
+  ...
+  extend-ignore = E501
+  extend-select = B901
```
cooperlees, line 203: Should we recommend same selection of all non optionals or does it not work like that anymore?
```suggestion
  extend-select = B,B901
```

README.rst
```diff
@@ -189,41 +192,56 @@ on the first line and urls or paths that are on their own line::
...
+
+Some checks might need other flake8 checks disabled - e.g. E501 must be disabled for
+B950 to be hit.
```
cooperlees, line 206:
```suggestion
Some flake8 checks require other flake8 checks disabled 
- `E501` must be disabled for `B950` of bugbear to be hit.
```

README.rst
```diff
@@ -189,41 +192,56 @@ on the first line and urls or paths that are on their own line::
...
+If you'd like all optional warnings to be enabled for you (future proof your config!),
+say ``B9`` instead of ``B901``. You will need flake8 3.2+ for this feature.
+
+For flake8 versions older than 4.0, you will need to use the ``--select`` command-line
+option or ``select=`` option in your config file. As of flake8 3.0, this option is a
+whitelist (checks not listed are implicitly disabled), so you have to explicitly specify
+all checks you want enabled (e.g. ``select = C,E,F,W,B,B901``).
+
+The ``--extend-ignore`` command-line option and ``extend-ignore=`` config file option
+require flake8 3.6+. For older flake8 versions, the ``--ignore`` and ``ignore=`` options
```
cooperlees, line 217: Some wording + syntax fixes
- Also, lets use notation Pythonistas are more familiar with (think PEP420 versioning specifiers)
```suggestion
The `--extend-ignore` command-line option and `extend-ignore=` config file options
require flake8 >=3.6. For older flake8 versions, the `--ignore` and `ignore=` options
```

README.rst
```diff
@@ -189,41 +192,56 @@ on the first line and urls or paths that are on their own line::
...
+beside those already handled by ``flake8-bugbear``. Calls to these method will no longer
+raise a ``B008`` warning.
+
+For example::
+
+  [flake8]
+  max-line-length = 80
+  max-complexity = 12
+  ...
+  extend-immutable-calls = pathlib.Path, Path
```
cooperlees, line 244: Awesome, thanks. This makes it a lot more accessible.

bugbear.py
```diff
@@ -127,7 +127,7 @@ def add_options(optmanager):
             help="Skip B008 test for additional immutable calls.",
         )
 
-    @lru_cache()
+    @lru_cache()  # noqa: B019
```
cooperlees, line 130: So meta ...

bugbear.py
```diff
@@ -507,6 +522,33 @@ def check_for_b017(self, node):
...
+    def check_for_b019(self, node):
+        if (
+            len(node.decorator_list) == 0
+            or len(self.contexts) < 2
+            or not isinstance(self.contexts[-2].node, ast.ClassDef)
+        ):
+            return
+
+        # Preserve decorator order so we can get the lineno from the decorator node rather than
+        # the function node (this location definition changes in Python 3.8)
```
cooperlees, line 534: So we can clean this up when >=3.8 as minimum version, yes? If so maybe explicitly say that so I and other maintainers remember.

tests/test_bugbear.py
```diff
@@ -130,8 +131,9 @@ def test_b007(self):
     def test_b008_extended(self):
         filename = Path(__file__).absolute().parent / "b008_extended.py"
 
-        mock_options = Mock()
-        mock_options.extend_immutable_calls = ["fastapi.Depends", "fastapi.Query"]
+        mock_options = Namespace(
+            extend_immutable_calls=["fastapi.Depends", "fastapi.Query"]
+        )
```
cooperlees, line 136: Cleaner, thanks!

cooperlees (changes requested):
Looks great. Thanks. Just some minor things:

MD syntax only needs ` and not `` - e.g. `select=`

All other requests are inline.

README.rst
sco1: Not needed with `extend-select`, since we're registering Bugbear as a plugin `B` is added to the default selects.

So 
```
ignore = E501
select = C,E,F,W,B,B901
```
and
```
extend-ignore = E501
extend-select = B901
```

Should be equivalent, with the added bonus that `extend-ignore` doesn't wipe out the default ignores like `ignore` does.

README.rst
sco1: To clarify on the ticks, most everything else in the README uses \`\` rather than \`, should I unify everything to \`?

bugbear.py
sco1: It depends where we ultimately want the emitted error to point. This isn't as slick as the set intersection but does consistently point the error at the first offending decorator rather than the function's `def` line, or the line where the decorators start (Python <3.8).

cooperlees:
If 3.6 is a pain, since it's EOL'd cpython support, I'm happy to drop it here too.

sco1:
It's not a major pain, it's mostly me just doing a dumb thing and trying to use `sys.version_info > (3, 6)` to detect Python >= 3.7. It's fixed now but I can also just drop it.

cooperlees:
If you have it fixed lets do 1 more 3.6 capable release and drop it next release :) Thanks!

khrapovs:
@sco1 I am totally confused. The error message says "class methods", but the actual behavior is to throw an error when "instance methods" are cached. I looked into the tests and this is how I read them. What do I miss?

jpy-git:
Answered in #250

Julian:
I may be missing it just scanning the above, but can someone mention the thinking on:

> Should this be restricted to caches that are explicitly unbounded, rather than any use of a built-in cache with a class method? i.e. this would only raise for functools.cache and functools.lru_cache(maxsize=None).

My reading of the upstream issue also seems to agree with the above (that you're likely OK if you're not setting `maxsize=None` and not creating hundreds of thousands of the object the method is on). Just confirming then, was the decision to warn always (and not just for `maxsize=None`) explicit?

sco1:
Just some thoughts as a non-maintainer:

> that you're likely OK if you're not setting `maxsize=None` and not creating hundreds of thousands of the object the method is on

Maybe, but I'm not sure many are even aware of the potential pitfall. Not to get into the hypothetical weeds too much, but the point where you'd incur a non-trivial performance cost is going to vary between projects, if you even approach it at all; theoretically it could even be well below the default `maxsize=128` if your objects are large enough. Rather than bugbear trying to guess at what that line is, IMO it being the "very opinionated" checker allows it to just say "hey this could be a problem" and let the project make the decision. Maybe this is to far into "what-if" land and ends up being more annoying than practical.

FWIW, there's also a similar discussion happening over at Pylint: https://github.com/PyCQA/pylint/issues/5670. It looks like they're opting to only warn for `functools.cache` and `maxsize=None`.

PR #220, opened 2022-01-25
Add B020 check to find for-loop control variable overiding iter set

### LINKED ISSUE #219: Proposed Check: same variable name of list and interating item in for-loop
I've already placed a question [here](https://stackoverflow.com/questions/70827394/is-there-pep-rule-or-pre-commit-check-to-prevent-same-name-of-variable-and-list). Then found this project.

```
items = [1, 2, 3]
for items in items:
    continue
# items == 3
```
This typo accidentally happened to me and caused a bug.  
Would be nice to catch it with something like: `Bxxx for-loop with the same variable name of the list and interating item found.`

Korben11:
# Description
A simple typo of the for-loop control variable name `item => items` in loop `[items for items in items]` can lead to an unintended change of the `items` variable.  

Fixes #219

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -247,7 +247,7 @@ MIT
 Change Log
 ----------
 
-22.1.27
+Unreleased
 ~~~~~~~~~~
 
 * B020: ensure loop control variable doesn't overrides iterable it iterates (#220)
--- a/README.rst
+++ b/README.rst
@@ -132,6 +132,8 @@ data available in ``ex``.
 
 **B018**: Found useless expression. Either assign it to a variable or remove it.
 
+**B020**: Loop control variable overrides iter set.
+
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -132,7 +132,7 @@ data available in ``ex``.
 
 **B018**: Found useless expression. Either assign it to a variable or remove it.
 
-**B020**: Loop control variable overrides iter set.
+**B020**: Loop control variable overrides iterable it iterates
 
 
 Opinionated warnings
--- a/README.rst
+++ b/README.rst
@@ -247,6 +247,11 @@ MIT
 Change Log
 ----------
 
+22.1.27
+~~~~~~~~~~
+
+* B020: ensure loop control variable doesn't overrides iterable it iterates (#220)
+
 22.1.11
 ~~~~~~~~~~
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -333,6 +333,7 @@ class BugBearVisitor(ast.NodeVisitor):
 
     def visit_For(self, node):
         self.check_for_b007(node)
+        self.check_for_b020(node)
         self.generic_visit(node)
 
     def visit_Assert(self, node):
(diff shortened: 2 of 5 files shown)


README.rst
```diff
@@ -132,6 +132,8 @@ data available in ``ex``.
 
 **B018**: Found useless expression. Either assign it to a variable or remove it.
 
+**B020**: Loop control variable overrides iter set.
```
cooperlees, line 135:
```suggestion
**B020**: Loop control variable overrides iterable it iterates
```
Lets be verbose here so people can't misinterpret

bugbear.py
```diff
@@ -871,6 +886,9 @@ def visit(self, node):
         "B018 Found useless expression. Either assign it to a variable or remove it."
     )
 )
+B020 = Error(
+    message="B020 Found for-loop with each item variable reassigning set variable."
```
cooperlees, line 890:
```suggestion
    message="B020 Found for loop that reassigns the iterable it is iterating with each iterable value."
```

tests/b020.py
```diff
@@ -0,0 +1,17 @@
+items = [1, 2, 3]
```
cooperlees, line 1: Lets comment on where the errors are to help future readers like other tests
```suggestion
"""
Should emit:
B020 - on lines 8 and 21
"""

items = [1, 2, 3]
```

tests/test_bugbear.py
```diff
@@ -249,6 +250,18 @@ def test_b018_classes(self):
...
 
+    def test_b020(self):
+        filename = Path(__file__).absolute().parent / "b020.py"
+        bbc = BugBearChecker(filename=str(filename))
+        errors = list(bbc.run())
+        self.assertEqual(
+            errors,
+            self.errors(
+                B020(3, 4, vars=("items",)),
+                B020(16, 9, vars=("values",)),
```
cooperlees, line 261:
```suggestion
                B020(8, 4, vars=("items",)),
                B020(21, 9, vars=("values",)),
```

cooperlees (changes requested):
Awesome, thanks!

Just some wording and doc string suggestions.

Please a change log entry please (It's in the README file too).

tests/b020.py
```diff
@@ -0,0 +1,17 @@
...
+for item in items:
+    print(item)
+
+values = {"secret": 123}
+
+for key, value in values.items():
+    print(f"{key=}, {value=}")
+
+for key, values in values.items():
+    print(f"{key=}, {values=}")
```
cooperlees, line 17:
```suggestion
    print(f"{key}, {value}")

for key, values in values.items():
    print(f"{key}, {values}")
```
Are these '=' meant to do something I don't understand in Python?

Korben11: '=' was introduced in [python 3.8](https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging)  
So output would be: `key='secret', value=123`  
Guess better to remove it for backward compatibility 😅

bugbear.py
```diff
@@ -871,6 +886,9 @@ def visit(self, node):
         "B018 Found useless expression. Either assign it to a variable or remove it."
     )
 )
+B020 = Error(
+    message="B020 Found for loop that reassigns the iterable it is iterating with each iterable value."
```
cooperlees, line 890:
```suggestion
    message=(
        "B020 Found for loop that reassigns the iterable it is iterating "
        + "with each iterable value."
    )
```

cooperlees (changes requested):
Left a line to long - We failed a bugbear check on bugbear! haha

README.rst
```diff
@@ -245,6 +247,11 @@ MIT
 Change Log
 ----------
 
+22.1.27
```
cooperlees, line 250: Generally we put unreleased here but I'll fix when ever I get around to releasing.
```suggestion
Unreleased
```

cooperlees (approved):
LGTM - Will be a good addition as I really dislike this and is an easy typo to make ...

PR #223, opened 2022-02-24
Fix crash on long empty string

mikhail-melnik:
This PR fixes crash on long empty string.


cooperlees:
Thanks for submitting this. Just for my understanding, what's the practical usage of having such a long string of white space in your source code?

mikhail-melnik:
Hi @cooperlees, thanks for checking my PR! We have a test case in our source code containing a structured block of text (message of specific format) to parse.

PR #231, opened 2022-03-17
B022: No arguments passed to contextlib.suppress

### LINKED ISSUE #222: Detect when no arguments were passed to 'contextlib.suppress'
Wrong code that is not detected by Flake8:

```python
import contextlib

with contextlib.suppress():  
    raise ValueError
```
Where the Flake8 error must be raised:

```python
import contextlib

with contextlib.suppress():  # <-- ERROR: NO ARGUMENTS WERE PASSED!
    raise ValueError
```

How the correct code may look:

```python
import contextlib

with contextlib.suppress(ValueError):  
    raise ValueError
```

jpy-git:
Closes #222. 

This covers the case where `contextlib.suppress` is used
```
import contextlib

with contextlib.suppress():  
    raise ValueError
```

However, it does not currently cover the case where `suppress` is imported from `contextlib`
```
from contextlib import suppress

with suppress():  
    raise ValueError
```
Do you know of any existing flake8 helper functions/methods to tie the `suppress` function node to it's import? 
I don't want to just ignore the one word `suppress` without checking incase it's actually a user defined function.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -134,7 +134,7 @@ data available in ``ex``.
 
 **B020**: Loop control variable overrides iterable it iterates
 
-**B022**: B022 No arguments passed to `contextlib.suppress`.
+**B022**: No arguments passed to `contextlib.suppress`.
 No exceptions will be suppressed and therefore this context manager is redundant.
 
 
--- a/README.rst
+++ b/README.rst
@@ -134,6 +134,9 @@ data available in ``ex``.
 
 **B020**: Loop control variable overrides iterable it iterates
 
+**B022**: B022 No arguments passed to `contextlib.suppress`.
+No exceptions will be suppressed and therefore this context manager is redundant.
+
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -143,6 +143,8 @@ as a joined string rather than a docstring.
 
 **B022**: No arguments passed to `contextlib.suppress`.
 No exceptions will be suppressed and therefore this context manager is redundant.
+N.B. this rule currently does not flag `suppress` calls to avoid potential false
+positives due to similarly named user-defined functions.
 
 
 Opinionated warnings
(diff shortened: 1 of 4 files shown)


cooperlees:
You're to quick - Sorry - I just merged another PR. I can look at cleaning up here or if you beat me that would be awesome too :D

jpy-git:
> You're to quick - Sorry - I just merged another PR. I can look at cleaning up here or if you beat me that would be awesome too :D

hehe no worries, I've resolved them now 😄

Would be good to get your take on this PR currently only covering `contextlib.suppress` rather just `suppress` since we'd have to resolve scope to figure out that the `suppress` being called is definitely the one from `contextlib`?

cooperlees (approved):
All makes sense to me, and test show us working and handling `*` expansion which is good.

cooperlees:
> Would be good to get your take on this PR currently only covering `contextlib.suppress` rather just `suppress` since we'd have to resolve scope to figure out that the `suppress` being called is definitely the one from `contextlib`?

Ahhh, I didn't think of this. Yeah, I feel assuming with `suppress` would be to noisy and we shouldn't do that.

README.rst
```diff
@@ -141,6 +141,10 @@ garbage collection.
 **B021**: f-string used as docstring. This will be interpreted by python
 as a joined string rather than a docstring.
 
+**B022**: No arguments passed to `contextlib.suppress`.
+No exceptions will be suppressed and therefore this context manager is redundant.
```
cooperlees, line 145: Maybe we could note here our miss of `from contextlib import suppress` and the reason why - noise as we're sure other people call methods suppress ...

jpy-git:
> Ahhh, I didn't think of this. Yeah, I feel assuming with `suppress` would be to noisy and we shouldn't do that.

Yeah that was my thinking too, better to avoid false negatives 😄 

Maybe we initially keep the rule like this and then I can make a separate issue to find a way to track variable/function scope when traversing the AST as a future refinement (I think pyflakes has something similar for tracking if imports get used).

README.rst
jpy-git: Good shout, will add a note 👍

jpy-git: I've added that in now

PR #233, opened 2022-03-17
Remove redundant W503 inline ignores

jpy-git:
We were already ignoring W503 in `.flake8` since it conflicts with black. This PR cleans up the redundant inline ignores.

Also there was a note saying to keep `.flake8` and `setup.cfg` in sync so I've also updated `setup.cfg` as well 😄 

On a side note, I'm pretty sure we can get rid of `.flake8` since `flake8` can use `setup.cfg` fine when developing. Saves having to keep them in sync.


cooperlees (approved):
Awesome, Thanks for the cleanup!

PR #237, opened 2022-03-21
Clarify B019 description

### LINKED ISSUE #236: False positive for B019
After the latest update, there are false positives for B019 on regular methods:
image

jpy-git:
Closes #236.
The previous description warned about using lru_cache on `class methods` which is redundant as `methods` belong to a class by definition and leads to confusion with `classmethods` (i.e. methods decorated with `@classmethod`).
﻿

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -132,7 +132,7 @@ data available in ``ex``.
 
 **B018**: Found useless expression. Either assign it to a variable or remove it.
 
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Rightio - Sure, this makes sense to me and helpful in future releases.

PR #238, opened 2022-03-22
B020: Fix comprehension false postives

### LINKED ISSUE #235: BUG/QUESTION: Not sure if B020 is expected to trigger here
running flake8 with the latest version of bugbear on this snippet
```python
for var in (var for var in range(10) if var % 2):
    print(var)
```
results in
```
t.py:1:5: B020 Found for loop that reassigns the iterable it is iterating with each iterable value.
```
Is this expected ? I don't understand how the error message relates to what's actually happening here, it seems to me that this could be a mistake.

jpy-git:
Closes #235.

Fixes a false positive for B020 where a variable defined in a comprehension is reused in the `for` loop targets.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -557,7 +557,7 @@ class BugBearVisitor(ast.NodeVisitor):
         targets.visit(node.target)
         ctrl_names = set(targets.names)
 
-        iterset = NameFinder()
+        iterset = B020NameFinder()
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Wow, thanks. So clean and Pythonic Objecty. Much love.

PR #239, opened 2022-03-22
B006 and B008: Cover additional test cases

### LINKED ISSUE #229: B008 fails for function calls within expressions
B008 does not detect cases where function calls are used in sub-expressions, like:

```python
import datetime as dt

def f(when=dt.datetime.now() + dt.timedelta(days=7)):
    pass
```

jpy-git:
Closes #229.
Detects function calls at any level of the default expression.
This includes instances of B008 which occur within instances of B006 (e.g. `def f(a=[datetime.now()])`) since a user may choose to turn one off but still want to be warned of the other.


.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
       - id: isort
 
   - repo: https://github.com/psf/black
-    rev: 21.10b0
+    rev: 22.1.0
```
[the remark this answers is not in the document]
cooperlees: > FYI, I've updated this as v22 introduced power operator hugging
> 
> so the list comprehensions kept alternating between `[i**2 for i in range(3)]` and `[i ** 2 for i in range(3)]` depending if I was running black via cli or pre-commit. Might as well update to avoid future confusion smile

Cool - let me open a issue to auto update these ...

cooperlees:
This is a big one. Calling in some much more ast savvy people to help review this one. There are some pieces I need to wrap my head around. Will eventually take the time and work it out, but adding others incase they can beat me.

bugbear.py
```diff
@@ -778,6 +740,79 @@ def visit(self, node):
...
+                # NOTE: regex derived from documentation at:
+                # https://docs.python.org/3/library/functions.html#float
+                inf_nan_regex = r"^[+-]?(inf|infinity|nan)$"
+                re_result = re.search(inf_nan_regex, str_val.lower())
+                is_float_literal = re_result is not None
+            else:
+                is_float_literal = False
+
+            if not is_float_literal:
+                self.errors.append(B008(node.lineno, node.col_offset))
```
Zac-HD, line 800: I know this was previously existing code, but we can shorten it *considerably* by using `ast.literal_eval()`:

```suggestion
                try:
                    value = float(ast.literal_eval(node.args[0]))
                except Exception:
                    pass
                else:
                    if math.isfinite(value):
                        self.errors.append(B008(node.lineno, node.col_offset))
```

(OK, I admit that this no longer warns on a redundant `float()` call wrapped around an "infinity literal" like `1e999`, but IMO that's fine.)

Zac-HD (approved):
Looks good to me!

bugbear.py
jpy-git: Thanks! I've updated the code accordingly and it looks a lot nicer now 😄

tests/b006_b008.py
```diff
@@ -1,18 +1,22 @@
...
 
 
+# B006
+# Allow immutable literals/calls/comprehensions
 def this_is_okay(value=(1, 2, 3)):
     ...
 
 
-def and_this_also(value=tuple()):
+async def and_this_also(value=tuple()):
```
cooperlees, line 19: Nice catch to add coverage

tests/b006_b008.py
```diff
@@ -42,35 +73,61 @@ def this_too(value=collections.OrderedDict()):
...
     ...
 
 
-def kwonlyargs_immutable(*, value=()):
-    ...
+# Recommended approach for function calls as default args
+LOGGER = logging.getLogger(__name__)
 
 
-def kwonlyargs_mutable(*, value=[]):
```
cooperlees, line 70: Do we need an empty list for coverage? I guess you removed this cause of all the list comprehension tests now ...

cooperlees (approved):
All looks good to me, just some questions for my interest.

Especially if Zac can't see anything wrong + his awesome suggestion.

tests/b006_b008.py
```diff
@@ -42,35 +73,61 @@ def this_too(value=collections.OrderedDict()):
...
+
+def dict_comprehension_also_not_okay(default={i: i**2 for i in range(3)}):
+    pass
+
+
+def set_comprehension_also_not_okay(default={i**2 for i in range(3)}):
+    pass
+
+
+def kwonlyargs_mutable(*, value=[]):
```
jpy-git, line 97: @cooperlees Re: the question on removing this. I rearranged this test file to structure it a bit more and help come up with extra test cases. The function is still here 😄

PR #243, opened 2022-03-25
Fix B008 edge case with lambda functions

### LINKED ISSUE #242: function calls inside lambda is warned as B008
Since flake8-bugbear 22.3.23, B008 warns the code below , but the function calls inside lambda is not called at function definition time.  It seems #239 caused this regression.
```python
def foo(f=lambda x: print(x)):
    f(1)


foo()
```

jpy-git:
Closes #242

In #239 I introduced the ability to detect functions at any depth of function default args to prevent things like:
```python
def foo(a=[datetime.now()], b=bar(datetime.now())): ...
```

One thing I overlooked was lambda functions which are slightly different given that their contents act as function body and are therefore only evaluated when the lambda function is called, therefore B008 doesn't apply to these.
```python
from datetime import datetime
import time

def foo(f=lambda x: datetime.now()):
    print(f(1))
    time.sleep(5)
    print(f(1))

foo()
```
prints
```
2022-03-25 11:36:22.999143
2022-03-25 11:36:28.004194
```

Fix is pretty simple, don't check inside lambda functions.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -791,6 +791,11 @@ class FuntionDefDefaultsVisitor(ast.NodeVisitor):
         # Check for nested functions.
         self.generic_visit(node)
 
+    def visit_Lambda(self, node):
+        # Don't recurse into lambda expressions
+        # as they are evaluated at call time.
+        pass
+
     def visit(self, node):
         """Like super-visit but supports iteration over lists."""
         self.arg_depth += 1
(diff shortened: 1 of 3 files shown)


cooperlees:
Wonder why CI didn't run here? That is very strange ...

jpy-git:
@cooperlees Strange, I've found that closing and reopening a PR seems to fix these things 😅

cooperlees:
And it did. Wonder where we can log the bug?

cooperlees (approved):
Makes sense to me. Thanks!

jpy-git:
> And it did. Wonder where we can log the bug?

Maybe a GitHub bug report: https://support.github.com/contact/bug-report

Although tbh I've only ever had it happen once before and closing/reopening or pushing another commit seems to re-trigger things so not too big an issue 😄

PR #246, opened 2022-03-28
Fix B010 lambda false positive

### LINKED ISSUE #122: B010 (setattr with constant value) false-positive with lambdas
I have this snippet of code:

```python
class FakeCookieStore:

    def __init__(self, has_setter):
        self.cookie_filter = None
        if has_setter:
            self.setCookieFilter = (lambda func:
                                    setattr(self, 'cookie_filter', func))
```

where flake8-bugbear complains:

```
test.py:7:37: B010 Do not call setattr with a constant attribute value, it is not any safer than normal property access.
```

However, using `self.cookie_filter = func` inside a lambda doesn't work.

My aim here is to have a test fake which either has a setter method or doesn't, to simulate different versions of a dependency.

jpy-git:
Closes #122. Allows use of `settatr` within lambda expressions since assignment operators aren't valid in this context.

### DIFF
--- a/tests/b009_b010.py
+++ b/tests/b009_b010.py
@@ -1,6 +1,6 @@
 """
 Should emit:
-B009 - Line 17, 18, 19
+B009 - Line 17, 18, 19, 44
 B010 - Line 28, 29, 30
 """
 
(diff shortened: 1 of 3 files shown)


tests/b009_b010.py
```diff
@@ -28,3 +28,20 @@
 setattr(foo, "bar", None)
 setattr(foo, "_123abc", None)
 setattr(foo, "abc123", None)
+
+# Allow use of setattr within lambda expression
+# since assignment is not valid in this context.
+c = lambda x: setattr(x, "some_attr", 1)
+
+
+class FakeCookieStore:
```
cooperlees, line 37: Now I want a cookie.

cooperlees (approved):
Nice, clean fix to avoid this edge case ...

PR #251, opened 2022-03-30
Ignore black formatting for b013 test case

jpy-git:
Should unblock #247


cooperlees (approved):
Thanks! Beat me. As usual.

PR #257, opened 2022-05-02
Fix missing space typos in B021, B022 error messages

gschaffner:


cooperlees (approved):
Thanks for spotting this - Wonder if we could link this ensuring message=() strs all but the last one have ' ' in them ...

#clowntownReviewsFromMe :P

PR #258, opened 2022-05-16
Add new check which finds duplicate except clauses

### LINKED ISSUE #254: New check which find duplicate except clauses
flake8-bugbear already find duplicate exeptions in the same tuple, but across different except's this is not done

**Examples**:
```python
try:
    pass
except ValueError:
    pass
except ValueError:
    pass
```
```python
try:
    pass
except (ValueError, TypeError):
    pass
except (ValueError, UnicodeError):
    pass
```

I'm happy tp contribute this check 😄

kasium:
Closes #254

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -150,6 +150,11 @@ No exceptions will be suppressed and therefore this context manager is redundant
 N.B. this rule currently does not flag `suppress` calls to avoid potential false
 positives due to similarly named user-defined functions.
 
+**B023**: Try with duplicate exceptions caught.
+If a exception is specified multiple times in the same or other except clauses in the same
+try-except statement, only the first specification is considered.
+All others can be removed.
+
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
--- a/bugbear.py
+++ b/bugbear.py
@@ -375,6 +375,7 @@ class BugBearVisitor(ast.NodeVisitor):
 
     def visit_Try(self, node):
         self.check_for_b012(node)
+        self.check_for_b023(node)
         self.generic_visit(node)
 
     def visit_Compare(self, node):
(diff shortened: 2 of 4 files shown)


kasium:
@cooperlees I just realized that B014 does partially the same job. The difference is, that it doesn't consider multiple except clauses. I would suggest to tune B014 to also consider multiple except's. This would mean that the error message would change. Is that fine?

cooperlees:
I think since *B014* looks for duplication per exception clause I'm a fan of the more explicit error message of the new *B023* stating that it's a redundant clause in a new except. Let me take a look at the wording to ensure this is explicit as it can be.

Thanks for asking.

README.rst
```diff
@@ -150,6 +150,11 @@ No exceptions will be suppressed and therefore this context manager is redundant
 N.B. this rule currently does not flag `suppress` calls to avoid potential false
 positives due to similarly named user-defined functions.
 
+**B023**: Try with duplicate exceptions caught.
+If a exception is specified multiple times in the same or other except clauses in the same
+try-except statement, only the first specification is considered.
+All others can be removed.
+
```
cooperlees, line 157:
```suggestion
**B023**: ``try-except`` block with duplicate exceptions found.
This check identifies exception types that are specified in multiple ``except`` 
clauses. The first specification is the only one ever considered, so all others can be removed.
```

Lets try and be explicit as we can here so people can understand how to fix it.

bugbear.py
```diff
@@ -1034,6 +1049,12 @@ def visit_comprehension(self, node):
         "context manager is redundant."
     )
 )
+B023 = Error(
+    message=(
+        "B023 Exception `{0}` is catched multiple times. "
```
cooperlees, line 1054:
```suggestion
        "B023 Exception `{0}` has been caught multiple times. "
```

cooperlees (changes requested):
Looks good - but tests show we've hit B014 being triggered at the same time. I feel this would be annoying for users. So I feel we either extend B014 or if B023 is hit, clear B014?
- OR, do people like the two explicit errors?

I personally think getting the two lints, since they are so similar annoying - So maybe combining would be preferential?

Also, some minor suggestions in the error output + README.

kasium:
I'm very sorry for the late pickup. Since this is heavily outdated, let me open a new PR

Issue #256, opened 2022-04-26
Possible B020 false positive for iteration with lambda


horpto:
```
class IndexDashboard(Dashboard):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)

        for app in sorted(apps.get_app_configs(), key=lambda app: app.name): # here is error
            self.children.append(
                modules.AppList(
                    title=app.verbose_name,
                    column=1,
                    collapsible=True,
                    models=(f"{app.name}.*",),
                    css_classes=("grp-closed",),
                ),
            )
```

As you can see there is no error actually. app in lambda function is local to lambda, not for __init__

NewGlad:
@cooperlees 
I think this issue can be closed since fix is merged

PR #259, opened 2022-05-16
Ignore lambda arguments for B020

NewGlad:
#256


cooperlees (changes requested):
Seems this has caused line 21 in b020.py to not raise the error anymore ... I'd like that to still trigger the error ...

cooperlees (approved):
LGTM. Thanks!

I’ll happily add to changes once i cut a release in next few weeks.

PR #265, opened 2022-06-23
Implement late-binding loop check

### LINKED ISSUE #263: Check for late-binding closures (e.g. lambdas) in loops
We just got bitten by the classic [late-binding of loop variables used in lambdas](https://docs.python-guide.org/writing/gotchas/#late-binding-closures) at work, and I'd like to make sure this doesn't happen again.

Specifically, it should be an error to use any loop variable inside the body of a function (whether `lambda ...` or `def ...`) defined inside that loop, unless the same name is a parameter of that function.  e.g. `for x in ...: fns.append(lambda x=x: x**2)` would be OK, but `lambda: x**2` would not.

Loops include (async) for loops, and list/set/dict/generator comprehensions; but exclude while-loops as they have no clearly associated loop variables.  While this can still be buggy, I'd like to keep the false-alarm rate as low as reasonably possible.

Related: https://github.com/PyCQA/flake8-bugbear/issues/79, https://github.com/PyCQA/flake8-bugbear/issues/175, and the much noisier [wemake-style-guide `LambdaInsideLoopViolation`](https://wemake-python-stylegui.de/en/latest/pages/usage/violations/best_practices.html#wemake_python_styleguide.violations.best_practices.LambdaInsideLoopViolation).

Zac-HD:
Closes #263.

Will write tests tomorrow; just wanted to get a proof of concept up tonight.


Zac-HD:
Ping @cooperlees?

cooperlees (approved):
Sorry about the delay. Been busy.

Makes sense to me. Includes Carl's suggestion + well documented + tested
- Plus links to a great doc for people to read to fully understand (like I did)

I feel this is going to find a lot of cases doing this. So lets release this on it's own and wreak havoc.

Zac-HD:
No worries, life happens!  Bring on the havoc, we'll see if people are complaining about false-alarms or just alarms they don't like 😁

Issue #273, opened 2022-08-16
Check for abstract base class definitions with no abstractmethods


Zac-HD:
```python
from abc import ABC, ABCMeta, abstractmethod

class Base(ABC):  # I'd like a lint error here if there are no @abstract* decorated methods

    # @abstractmethod
    def method(self):
        ...

class MyABC(metaclass=ABCMeta):  # and here too for the same reason
    pass
```

Suggested message: `{classname} is an abstract base class, but it has no abstract methods.  Remember to use @abstractmethod, @abstractclassmethod, and/or @abstractproperty decorators.`

We might also want to emit a warning for a probably-missing-decorator on any method of an ABC where the whole method body is `...`, even if there are some abstract-decorated methods on the class.

cooperlees:
I can't see how this could draw false positives ... so can do nothing but help. I feel this fits things bugbear looks for ... PR will be accepted.

jakkdl:
I'm starting on a PR now!

> We might also want to emit a warning for a probably-missing-decorator on any method of an ABC where the whole method body is ..., even if there are some abstract-decorated methods on the class.
How about extending this to the body being anything but `pass`, `...` or string constants - to also catch
```python
# @abstractmethod
def method(self):
  """my abstract method"""
  pass
```

Zac-HD:
Closed by #274.

jakkdl:
If you also want
> We might also want to emit a warning for a probably-missing-decorator on any method of an ABC where the whole method body is ..., even if there are some abstract-decorated methods on the class.

you can open a separate issue for it, should be an easy one as well :)

Zac-HD:
Nah,  let's release what we've got now and let future features take care of themselves.

(though Cooper,  you might want to copy the auto release set up from flake8-async to make releases really easy in future?)

cooperlees:
I'm not a fan of that tbh as we don't have good enough change log enforcement etc. with this project so the 1 commit + release via GitHub is there enough for me imo.

PR #274, opened 2022-08-17
Add b024: abstract class with no abstract methods

jakkdl:
First requested error in #273

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,6 +154,8 @@ positives due to similarly named user-defined functions.
 the loop, because `late-binding closures are a classic gotcha
 <https://docs.python-guide.org/writing/gotchas/#late-binding-closures>`__.
 
+**B024**: Abstract base class with no abstract method. Remember to use @abstractmethod, @abstractclassmethod, and/or @abstractproperty decorators.
+
 Opinionated warnings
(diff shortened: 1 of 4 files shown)


jakkdl:
Couple possibilities on how liberal the name checks should be to balance false/missed alarms. You can check `tests/b024.py` for where the line is drawn currently.

@Zac-HD

Zac-HD:
I'd warn only if the name is exactly `ABC` or `abc.ABC`, missing the alarm if you `[from ...] import ... as` seems fine to me.

jakkdl:
I was thinking more if any other libraries implement `ABC`'s, but I guess that's very rarely the case (and it that case they might be doing fancy stuff such that this would be a false alarm anyway).

(Previous CI run failed due to me forgetting to update lines after black formatted the test file)

cooperlees:
Hmm - Where be the CI? I clicked approve :|

Zac-HD:
Stuck in queued, and there's an incident https://www.githubstatus.com/incidents/8gpymn7rystj

Should be fine in a few hours?

Zac-HD (approved):
Looks great - thanks @jakkdl!

PR #287, opened 2022-09-20
B026 - Argument unpacking after keyword argument

### LINKED ISSUE #286: Proposed check: don't allow argument unpacking in a call after a keyword argument
Imagine a function like

```py
def foo(bar, baz):
    pass
```

The following call is a syntactically incorrect:

```py
foo(baz="baz", "bar")
```

However, with unpacking the following is allowed:

```py
foo(baz="baz", *["bar"])
```

[However](https://github.com/python/cpython/issues/82741#issuecomment-1093845041),

> [T]he resulting argument binding is awkward, and almost never does what you want/expect it to

It was [not ruled a bug or a behavior that can or should be deprecated](https://github.com/python/cpython/issues/82741#issuecomment-1093845049). Still, I think in most cases this is probably an oversight and should not be used.

If we want this check, I would be willing to contribute a patch.

pmeier:
Closes #286.


bugbear.py
```diff
@@ -641,6 +643,19 @@ def is_abstract_decorator(expr):
...
+        try:
+            starred = next(arg for arg in call.args if isinstance(arg, ast.Starred))
+        except StopIteration:
+            return
+
+        if any(
+            (keyword.value.lineno, keyword.value.col_offset)
+            < (starred.lineno, starred.col_offset)
+            for keyword in call.keywords
+        ):
```
pmeier, line 656: I don't know if `call.keywords` is ordered by occurrence in the definition. If yes, we can also get away with only checking the first element

```suggestion
        if (call.keywords[0].value.lineno, call.keywords[0].value.col_offset) < (starred.lineno, starred.col_offset):
```

bugbear.py
```diff
@@ -641,6 +643,19 @@ def is_abstract_decorator(expr):
 
         self.errors.append(B024(node.lineno, node.col_offset, vars=(node.name,)))
 
+    def check_for_b026(self, call: ast.Call):
+        try:
```
pmeier, line 647: This guard is good in general, but especially needed if we go with the suggestion above to avoid catching the empty keywords later on.

```suggestion
        if not call.keywords:
            return
        try:
```

tests/b026.py
```diff
@@ -0,0 +1,16 @@
...
+def foo(bar, baz, bam):
+    pass
+
+
+foo("bar", "baz", bam="bam")
+foo("bar", baz="baz", bam="bam")
+foo(bar="bar", baz="baz", bam="bam")
+foo(bam="bam", *["bar", "baz"])
+foo(baz="baz", bam="bam", *["bar"])
+foo(bar="bar", baz="baz", bam="bam", *[])
```
cooperlees, line 16:
```suggestion
bar_bar = ["bar", "baz"]
foo("bar", "baz", bam="bam")
foo("bar", baz="baz", bam="bam")
foo(bar="bar", baz="baz", bam="bam")
foo(bam="bam", *["bar", "baz"])
foo(bam="bam", *bar_baz)
foo(baz="baz", bam="bam", *["bar"])
foo(bar="bar", baz="baz", bam="bam", *[])
```

Don't know if testing with a variable name adds an edge case here but if it does can we add :D ...

cooperlees (commented):
All looks good to me. I would love a second review from an AST stronger person than I am. Maybe adding my test suggestion might cover another edge case we should support but I am not sure ... 

*(yes it's ironic that someone so AST challenged maintains this project)*

pmeier:
> I would love a second review from an AST stronger person than I am

Any suggestions?

tests/b026.py
Zac-HD: The logic we actually have doesn't depend on using constants vs variable names here, but I'm generally in favor of more-comprehensive testing as a guard against regressions 👍

Zac-HD:
This looks pretty good to me, with the substantial caveat that function calls can have *multiple* unpackings and you're only checking the first:

```python
def f(a, b, c, d):
    print(a, b, c, d)

f(
    *[1],  # currently no lint error, because this is before the keyword and we stop at the first
    d=0,
    *[2],  # want a lint here
    *[3],  # and also here
)
# prints: 1 2 3 0
```

@cooperlees I assume you know about [greentreesnakes docs](https://greentreesnakes.readthedocs.io/en/latest/nodes.html#Call)?  They make me look better at AST stuff than I deserve 😉

pmeier:
> substantial caveat that function calls can have _multiple_ unpackings and you're only checking the firs

Good catch!

tests/b026.py
pmeier: I've kept it.

bugbear.py
pmeier: I've checked against all tests in this repo and the keywords are indeed ordered by occurrence.

pmeier:
Sorry for the weird review requests. GH doesn't let me do it properly.

Zac-HD:
Implementation looks great to me - thanks!

Before merging I'd also want a longer/more informative message, along the lines of 'To avoid confusing or misleading calls, put star-args before keyword arguments so that the order of arguments matches the order of the parameters.'. Not delighted by that wording, but I think it's important to say *why* as well as *what* to help the user learn.  And better to comment now than come back with something better a week late 😅

README.rst
```diff
@@ -160,6 +160,8 @@ the loop, because `late-binding closures are a classic gotcha
 This check identifies exception types that are specified in multiple ``except``
 clauses. The first specification is the only one ever considered, so all others can be removed.
 
+**B026**: Argument unpacking after keyword argument.
```
cooperlees, line 163:
```suggestion
**B026**: Argument unpacking after keyword argument is strongly discouraged. This behavior is uncommon and can lead to unnecessary confusion when reading the code thus making it prone to causing bugs. There was cpython discussion of disallowing but legacy usage and parser limitations make it difficult.
```
+1 on more description here. A suggestion that I'm more than happy to have reworded + made better.

cooperlees (changes requested):

README.rst
Zac-HD:
```suggestion
**B026**: Star-arg unpacking after a keyword argument is strongly discouraged, because
it only works when the keyword parameter is declared after all parameters supplied by 
the unpacked sequence, and this change of ordering can surprise and mislead readers.
```

I think this is compact enough to also use as the runtime message, and explains specifically what we think is wrong (the mismatched order can be misleading) rather than non-actionable historical details.

I do like the historical details and I'd be happy to keep "There was [cpython discussion of disallowing](https://github.com/python/cpython/issues/82741) this syntax, but legacy usage and parser limitations make it difficult." with that link in the readme too, but not print it at runtime.

pmeier: I'm not super happy about the wording and needed to read it multiple times to understand that it indeed correctly explains the issue. Still, I have no better way so what gives. This is really hard to explain without an example TBH.

+1 on keeping the historical details to the README.

cooperlees (approved):
Awesome. All LGTM.

cooperlees:
Will work on https://github.com/PyCQA/flake8-bugbear/pull/288 and release after we merge that.

PR #288, opened 2022-09-22
Migrate metadata from setup.py to setup.cfg

cclauss:
This pull request was created using https://pypi.org/project/setuptools-py2cfg plus manual modifications.

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use
> We recommend users expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (or even omit it completely if applicable).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,57 @@
+[metadata]
+license_file = LICENSE
+name = flake8-bugbear
+version = attr: bugbear.__version__
+author = Łukasz Langa
+author_email = lukasz@langa.pl
+license = MIT
+description = A plugin for flake8 finding likely bugs and design problems in your program. Contains warnings that don't belong in pyflakes and pycodestyle.
+keywords =
+    flake8
+    bugbear
+    bugs
+    pyflakes
+    pylint
+    linter
+    qa
+url = https://github.com/PyCQA/flake8-bugbear
+long_description = file: README.rst
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Environment :: Console
+    Framework :: Flake8
+    Intended Audience :: Developers
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,10 +1,61 @@
+[metadata]
+license_file = LICENSE
+name = flake8-bugbear
+version = attr: bugbear.__version__
```
cooperlees, line 4: TIL - This will help releases. Thanks.

cooperlees (approved):
LGTM - Thanks ...

Just one question; why not move to `pyproject.toml`?
  - I'm happy to take this step than pyproject.toml but just wondering why
  - I believe setuptools support enough of the PEP now for pyproject.toml for pure python projects like us - I could be wrong tho

cclauss:
> why not move to pyproject.toml? 

Because flake8 does not support py project.toml.  PyCQA/flake8#234

cooperlees:
Good enough reason for now. Thanks for this. Lets roll from here and move to `pyproject.toml` when flake8 does.

wwuck:
@cooperlees @cclauss FYI, there is a plugin that allows `pyproject.toml` configuration for flake8.

https://pypi.org/project/Flake8-pyproject/

cclauss:
My sense is that a separate, supported text file is safer in the near-term and long-term than adding another dependency.

In fact, there are numerous outside dependencies that attempt to paper over this situation:
* https://pypi.org/project/Flake8-pyproject
* https://pypi.org/project/pflake8
* https://pypi.org/project/flake9
* https://pypi.org/project/flakehell
* https://pypi.org/project/ruff <-- Shows great promise

But none of these are from PyCQA and I believe that the best approach would be for the PyCQA to _solve_ this problem internally rather than encouraging the community to paper over it.

cooperlees:
Totally agree with PyCQA solving this and making it first class.

Lets follow + comment on https://github.com/PyCQA/flake8/issues/234

PR #291, opened 2022-09-23
PEP 621: Migrate to pyproject.toml

cclauss:
Migrate to `pyproject.toml` using [`ini2toml`](https://pypi.org/project/ini2toml) to do the file conversion and running [`validate-pyproject`](https://github.com/abravalheri/validate-pyproject) in `pre-commit` to validate the results.
* Comment out `tool.setuptools.test-suite` which was deprecated in `setuptool` v41.5.0
* Currently `flake8` is not compatible with `pyproject.toml`
* Drop Python 3.6 and add Python 3.11 release candidate 2 -- https://www.python.org/download/pre-releases


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", '3.11-dev']
```
cooperlees, line 11:
```suggestion
        python-version: ["3.7", "3.8", "3.9", "3.10", '3.11-dev - 3.11']
```
I think this will make us auto jump to 3.11 when it's out. `"3.11.0-rc - 3.11"` go suggested to me for black's CI

.github/workflows/pypi_upload.yml
```diff
@@ -14,6 +14,8 @@ jobs:
 
     - name: Set up Python
       uses: actions/setup-python@v4
+      with:
+        python-version: 3.x
```
cooperlees, line 18: Doesn't this default to that or did I break it? Happy to be more explicit here too.

setup.cfg
```diff
@@ -1,57 +1,3 @@
...
-flake8.extension = B = bugbear:BugBearChecker
-
-[options.extras_require]
-dev =
-    coverage
-    hypothesis
-    hypothesmith>=0.2
-    pre-commit
-
 [flake8]
```
cooperlees, line 1: Can't wait for flake8 to support config in the toml

cooperlees (approved):
LGTM - Thanks!
- I checked  - flake8 is now >= 3.7 so happy to drop 3.6 finally.

Will leave up for at least 24 hours for some others to potentially comment.

.github/workflows/pypi_upload.yml
cclauss: It used to be REQUIRED but perhaps they _fixed_ it after user complaints.  If your mods work then it is now optional.

cooperlees: https://github.com/pypa/bandersnatch/blob/main/.github/workflows/pypi_upload.yml#L18 - Ahh seems I did * here to get latest available python 3

setup.cfg
cclauss: My sense is that will take time...

.github/workflows/pypi_upload.yml
cclauss: That would actually autoupgrade to Python 4!!!  You _are_ brave.  ;-)

cooperlees: Ya, I roll forward and just fix when things break everywhere ... rather than pin and forget to move things like most people seem to prefer to do which I don't get ...

PR #292, opened 2022-09-23
Add Python 3.11 release candidate 2 to the testing

cclauss:
Python 3.11 will be on average 22% faster than Python 3.10 so let’s give it a spin...
https://www.python.org/download/pre-releases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11.0-rc - 3.11"]
         os: [ubuntu-latest]
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
```
cooperlees, line 11:
```suggestion
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev - 3.11"]
```
Can we see if this works - It was suggested to me on black's CI.

cooperlees (approved):
3.11 is going to be awesome.

.github/workflows/ci.yml
cclauss:
```suggestion
        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11.0-rc - 3.11"]
```

cooperlees (approved):

PR #294, opened 2022-10-03
add tox config

jakkdl:
I wrote a basic [tox](https://tox.wiki/en/latest/index.html) config when writing #281 as I was fighting with various version differences and kept failing the CI, and figured other people might find it useful as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,3 +1,4 @@
+# Uses https://github.com/ymyzk/tox-gh-actions
 name: ci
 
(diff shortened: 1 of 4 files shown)


cooperlees (approved):
Cool, as much as I dislike too, I like it's helpfulness + uiquotious nature in the python OSS ecosystem so happy to use it here.

If you have time - Want to change our GitHub action to use it to so we know it's working? If not, I can eventually get to it.

cooperlees (approved):
Thanks. Let's move to `tox` life.

PR #298, opened 2022-10-17
Make bugbear slightly more polite

sam-w:
Bugbear is great at catching bugs in our code, but can come across a bit rude when commenting on pull requests 😂  Obviously it's a robot, but we like to encourage polite discourse

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1213,7 +1213,7 @@ B017 = Error(
 )
 B018 = Error(
     message=(
-        "B018 Found useless expression. Consider either assigning it to a"
+        "B018 Found useless expression. Consider either assigning it to a "
         "variable or removing it."
     )
 )
--- a/bugbear.py
+++ b/bugbear.py
@@ -1192,8 +1192,8 @@ B014.redundant_exceptions = {
 }
 B015 = Error(
     message=(
-        "B015 Pointless comparison. This comparison does nothing but waste "
-        "CPU instructions. Either prepend `assert` or remove it."
+        "B015 Result of comparison is not used. This line doesn't do"
+        "anything. Did you intend to prepend it with assert?"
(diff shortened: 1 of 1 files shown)


bugbear.py
```diff
@@ -1213,7 +1213,8 @@ def visit_Lambda(self, node):
 )
 B018 = Error(
     message=(
-        "B018 Found useless expression. Either assign it to a variable or remove it."
+        "B018 Found useless expression. Consider either assigning it to a"
```
cooperlees, line 1216:
```suggestion
        "B018 Found useless expression. Consider either assigning it to a "
```
Whitespace

bugbear.py
```diff
@@ -1192,8 +1192,8 @@ def visit_Lambda(self, node):
 }
 B015 = Error(
     message=(
-        "B015 Pointless comparison. This comparison does nothing but waste "
-        "CPU instructions. Either prepend `assert` or remove it."
```
cooperlees, line 1196: I personally feel the blunt nature of the message makes it clearer if what the check looks for.
The new message will make people think into it more and possibly over complicate their thoughts. But open to others thoughts …

cooperlees (approved):
Seems fair. I’d like others thoughts if anyone comes by …

bugbear.py
sam-w: I definitely see the tradeoff here, and agree that the cause of the error should be the first thing the user sees. The aim with the change of wording was to retain that, while softening the order to a suggestion.

cooperlees (approved):
Thanks! Sorry about being slow. I’m on holidays at the moment.

sam-w:
@cooperlees no probs! Thanks for accepting the change, hope your holidays were excellent

PR #303, opened 2022-10-25
Fix some B023 false alarms

### LINKED ISSUE #269: B023 false alarms
When upgrading, some false positives were found for B023 (implimented in https://github.com/PyCQA/flake8-bugbear/pull/265).

## Example

```python
"""B023.py"""
def filter_values(values: list[list[int]], max_percentage: float):
    for value in values:
        filter_val = max(value) * max_percentage
        yield list(filter(lambda x: x < filter_val, value))

```

```console
$ flake8 B023.py
bugbear_me.py:4:41: B023 Function definition does not bind loop variable 'filter_val' is valid, and doesn't fall into the 
```

A silly example here, but one where the use of `filter_val` is valid.

cc @Zac-HD
In the PR you mentioned some hard to detect false positives that you were okay with. Was this kind of pattern one of the ones you had in mind?

Thanks for this check btw, I think it's a good idea. I've fallen for this in the past!

jakkdl:
Picking up from https://github.com/PyCQA/flake8-bugbear/compare/main...Zac-HD:flake8-bugbear:B023-false-alarms
Partially fixes #269 

Fixes some common false alarms where a function is immediately consumed, as an argument to `filter` or `reduce`, as a keyword argument to `key=` and as a return value.

**TODO**: I'm not sure how to handle more complex `return` cases, see the last test cases in `b023.py`.

I'm also not sure if this solution is too generous, it currently marks all direct child nodes to calls to `filter`&`reduce` as safe - but @Zac-HD's [comment](https://github.com/PyCQA/flake8-bugbear/issues/269#issuecomment-1172975488) implies it should maybe be stricter:
> Unfortunately it's more difficult to avoid this false alarm than you might think: for example "lambdas are OK if they're the first argument in a call to filter()" would miss bugs where the filter iterable (and therefore lambda) isn't immediately evaluated.

I'm not super read up on the problem and didn't take time to fully understand all comments with false alarms in the original issue, so would love some eyes on this. But it fixes all test cases @Zac-HD added at least.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -280,7 +280,7 @@ class BugBearVisitor(ast.NodeVisitor):
             names = [_to_name_str(e) for e in node.type.elts]
             as_ = " as " + node.name if node.name is not None else ""
             if len(names) == 0:
-                vs = ("`except (){}:`".format(as_),)
+                vs = (f"`except (){as_}:`",)
                 self.errors.append(B001(node.lineno, node.col_offset, vars=vs))
             elif len(names) == 1:
                 self.errors.append(B013(node.lineno, node.col_offset, vars=names))
@@ -912,7 +912,7 @@ class BugBearVisitor(ast.NodeVisitor):
                     uniques.add(name)
                 seen.extend(uniques)
         # sort to have a deterministic output
-        duplicates = sorted(set(x for x in seen if seen.count(x) > 1))
+        duplicates = sorted({x for x in seen if seen.count(x) > 1})
         for duplicate in duplicates:
             self.errors.append(B025(node.lineno, node.col_offset, vars=(duplicate,)))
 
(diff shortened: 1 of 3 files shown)


Zac-HD (approved):
Looks good to me - thanks @jakkdl!

cc @cooperlees to merge?

cooperlees:
Thanks. Lots going on here I don't get but I'll trust you two! Tests look good.

Zac-HD:
> I'm also not sure if this solution is too generous, it currently marks all direct child nodes to calls to `filter`&`reduce` as safe - but @Zac-HD's [comment](https://github.com/PyCQA/flake8-bugbear/issues/269#issuecomment-1172975488) implies it should maybe be stricter:
> 
> > Unfortunately it's more difficult to avoid this false alarm than you might think: for example "lambdas are OK if they're the first argument in a call to filter()" would miss bugs where the filter iterable (and therefore lambda) isn't immediately evaluated.

In short, there was enough chatter in that thread to make it clear the false-alarm rate was unacceptably high, and if we miss some real alarms that's basically just the natural downside of minimally-resourced open source projects!  Better to catch most bugs without annoying people than have a tool that nobody uses...

PR #305, opened 2022-10-26
also fix map in B023

### LINKED ISSUE #269: B023 false alarms

jakkdl:
and updated some comments
fixes #269 
@Zac-HD

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -592,7 +592,7 @@ class BugBearVisitor(ast.NodeVisitor):
                 # check for filter&reduce
                 if (
                     isinstance(node.func, ast.Name)
-                    and node.func.id in ("filter", "reduce")
+                    and node.func.id in ("filter", "reduce", "map")
                 ) or (
                     isinstance(node.func, ast.Attribute)
                     and node.func.attr == "reduce"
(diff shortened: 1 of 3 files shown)


jakkdl:
I mixed up "delete merged branch" with "delete repository" ... I'll try to restore the repo but should be mergeable

jakkdl:
GitHub support restored my repository, so all's back to normal ^^;

PR #306, opened 2022-10-26
B027 ignore @overload decorator

### LINKED ISSUE #304: B027 false-positive on `@overload`ed methods
Running flake8 on the following code with flake8-bugbear 22.10.25:

```python
import abc
from typing import IO, Literal, overload

class Opener(abc.ABC):
    @overload
    def open(self, mode: Literal["r"] = "r") -> IO[str]:
        ...

    @overload
    def open(self, mode: Literal["rb"] = "rb") -> IO[bytes]:
        ...

    @abc.abstractmethod
    def open(self, mode: Literal["r", "rb"] = "r") -> IO:
        ...
```

results in reports of B027 violations for the `@overload`-decorated `open()` methods, which is wrong.

jakkdl:
fixes #304 
@Zac-HD

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -666,6 +666,14 @@ class BugBearVisitor(ast.NodeVisitor):
                 isinstance(expr, ast.Attribute) and expr.attr[:8] == "abstract"
             )
 
+        def is_overload(expr):
+            return (isinstance(expr, ast.Name) and expr.id == "overload") or (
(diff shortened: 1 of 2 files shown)


jakkdl:
WHOOPS

Zac-HD (approved):
Thanks for the quick fix!

PR #307, opened 2022-10-29
fix typo in unused comparison (B015) message

jameslamb:
Thanks so much for this awesome plugin! I just learned about it tonight and tried it out on one of my projects, and it already found a few bugs that `pylint` and standard `flake8` didn't 😍 

While doing that, I saw the following message.

> B015 Result of comparison is not used. This line doesn't doanything. Did you intend to prepend it with assert?

I think `"doanything"` was supposed to be `"do anything"`.

This PR proposes that change.

Thanks for your time and consideration.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1239,7 +1239,7 @@ B014.redundant_exceptions = {
 }
 B015 = Error(
     message=(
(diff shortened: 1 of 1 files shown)


cooperlees (approved):
Nice catch. Many thanks!

PR #309, opened 2022-11-07
B027 now ignores @[anything].overload to reduce false positives.

### LINKED ISSUE #304: B027 false-positive on `@overload`ed methods

jakkdl:
Fixes #304 

it also ignores `@a.b.c.d.overload` in case somebody is writing a custom overload or importing typing in some *very* weird way - but that seems more likely than disaster striking because the check didn't trigger imo. Didn't bother writing a test for that though

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -297,6 +297,10 @@ MIT
 Change Log
 ----------
 
+Future
+~~~~~~~~~
+* B027: ignore @overload when typing is import with other names
+
 22.10.27
 ~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Agree it's better to be less nosy here. Thanks for adding the tests too!

PR #314, opened 2022-12-06
add b905: require strict= argument to zip()

### LINKED ISSUE #311: Opinionated warning: require `strict=` argument to `zip()` builtin
In Python 3.10, [PEP-618 added optional length-checking to `zip()`](https://peps.python.org/pep-0618/).

Requiring uses of `zip()` to declare either `strict=` would be a nice option to prevent bugs when the argument was omitted, and the iterables were unexpectedly of different lengths.  Of course it could also be pretty noisy, not to mention requiring Python 3.10 or later, so I think this should be disabled by default.

jakkdl:
fixes #311

the test for b020 contains zip(), so rather than add b905 to it's error list I added a filter.
fixed a typo in changelog entry

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -200,6 +200,8 @@ or ``raise ... from None`` to distinguish them from errors in exception handling
 See `the exception chaining tutorial <https://docs.python.org/3/tutorial/errors.html#exception-chaining>`_
 for details.
 
+**B905**: `zip()` without an explicit `strict=` parameter. Added with python3.10, so don't enable this flag for code that should work on previous versions. https://peps.python.org/pep-0618/
+
 **B950**: Line too long. This is a pragmatic equivalent of
 ``pycodestyle``'s ``E501``: it considers "max-line-length" but only triggers
 when the value has been exceeded by **more than 10%**. You will no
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -200,6 +200,8 @@ or ``raise ... from None`` to distinguish them from errors in exception handling
 See `the exception chaining tutorial <https://docs.python.org/3/tutorial/errors.html#exception-chaining>`_
 for details.
 
+**B905**: `zip()` without an explicit `strict=` parameter. Added with python3.10, so don't enable this flag for code that should work on previous versions. https://peps.python.org/pep-0618/
```
cooperlees, line 203:
```suggestion
**B905**: `zip()` without an explicit `strict=` parameter set. `strict=True` causes the resulting iterator will raise a `ValueError` if the arguments are exhausted at differing lengths. This was added with python3.10 so don't enable this flag for code that should work on <3.10. For more information: https://peps.python.org/pep-0618/
```

cooperlees (approved):
Thanks. Looks good to me. Suggested a more verbose explanation of the error. Also, thanks for explaining the change to B020.

Zac-HD (approved):
Looks great, thanks @jakkdl (and @cooperlees for the review + readme update!) 🚀:tada:

Zac-HD:
@cooperlees do you want to handle the release?

cooperlees:
Yeah I can do later today

Zac-HD:
🎵 *I'm gonna have so many errors* 🎶 (it's great 😁)

Issue #267, opened 2022-07-01
B018 does not detect module level useless expressions


graingert:
with the following:

```python
def ham():
    pass


["hello", "world"]
```
when running `pipx run --spec="flake8-bugbear" flake8 fluke8.py --include="B018"`, I expected to see:
```
flake8.py:5:1: B018 Found useless expression. Either assign it to a variable or remove it.
```

r-downing:
ok to close?

cooperlees:
Thanks for noting.

PR #315, opened 2022-12-07
Enable module-level checks for B018

inwaves:
Hi!

This PR addresses #267 and enables B018 checks for module-level statements. I'm doing this by visiting the `Module` component in the AST first, and then any children. Let me know if I've missed anything.

Thanks!


cooperlees (approved):
Awesome. Many thanks! Module level unless assignments are very worth removing. Nice addition. This could be helpful in big modules.

PR #316, opened 2022-12-07
add B906: visit_ function with no calls to a visit function

### LINKED ISSUE #313: New check: visit_ function with no generic_visit/visit
This is kind of super niche, and mostly just of interest to flake8 plugin writers in particular, but I accidentally forgot to continue `visit`'ing in a couple functions I wrote recently in flake8-trio and realized it'd be incredibly handy - and very easy to implement - an optional check where if a function name matches `visit_*` it raises a warning if there's no `self.visit(...)` or `self.generic_visit(...)` within it's body.

I imagine it'd be nice to have that in the CI for flake8-bugbear itself as well, even though your structure makes it less likely than it is for flake8-trio currently.

jakkdl:
Fixes #313 

Also adds `#noqa` to the parts of bugbear's own code that explicitly breaks this.
Opted not to check that it's inside a class, that the parameters are standard, or that `ast` has been imported - which feels fine since it's an optional check.

@Zac-HD

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -205,6 +205,10 @@ to raise a ``ValueError`` if the arguments are exhausted at differing lengths. T
 was added in Python 3.10, so don't enable this flag for code that should work on <3.10.
 For more information: https://peps.python.org/pep-0618/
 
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Looks good to me - thanks!  @cooperlees to merge?

cooperlees (approved):
Awesome. Thanks - Now we need to get all flake8 AST based plugins to use bugbear!

Issue #300, opened 2022-10-24
Fix tox for python3.11 and future new versions


cooperlees:
tox isn't doing anything useful in 3.11 CI. Example: https://github.com/PyCQA/flake8-bugbear/actions/runs/3316961501/jobs/5479325116#step:5:41

```console
Run tox
  tox
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.0-rc.[2](https://github.com/PyCQA/flake8-bugbear/actions/runs/3316961501/jobs/5479325116#step:6:2)/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.0-rc.2/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.0-rc.2/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.0-rc.2/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.0-rc.2/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.0-rc.2/x64/lib
Warning: WARNING: tox-gh-actions couldn't find environments matching the provided factors from envlist. Please use `tox -vv` to get more detailed logs.
___________________________________ summary ____________________________________
  congratulations :)
```

Need to workout how to fix as I'm no tox guru. Help would be appreciated.

inwaves:
Hi! I just took a look at this. I think it's related to how `tox-gh-actions` parses the Python versions specified under `[gh-actions]`. As far as I can tell, it's only considering `major.minor`, which doesn't match our request for `major.minor.micro-releaselevel`. There are two options here:
- in the short term, we can change the version string to `3.11: py311`. See this [CI run](https://github.com/inwaves/flake8-bugbear/actions/runs/3635956234/jobs/6135511669#step:6:14) for confirmation.
- in the long term, we should try to get a PR in upstream so we can use release ranges and extended syntax, as spcified [here](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input).

inwaves:
Hm, I wonder if getting rid of the plugin is causing more trouble than leaving it in for now, and updating the 3.11 to just 3.11 (instead of rc). We might be able to get a change in upstream before 3.12.

[...]

PR #318, opened 2022-12-08
Move to official 3.11 CI

cooperlees:
Address part of #300

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools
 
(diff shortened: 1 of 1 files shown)


cooperlees:
This now works, but we run 3.10 and 3.11 here for some reason (I guess I'd cause 3.10 exists in the env ...)
```
py310: OK (26.61=setup[13.54]+cmd[12.92,0.15] seconds)
[57](https://github.com/PyCQA/flake8-bugbear/actions/runs/3645653834/jobs/6155984719#step:6:58)
  py311: OK (26.29=setup[10.88]+cmd[15.24,0.17] seconds)
```

This is better than where we were, so I think I'm going to merge this.

PR #321, opened 2022-12-12
Skip B905 automatically if python <3.10 is used

### LINKED ISSUE #320: B905 is raised on python 3.7
When having a zip call like ``zip(a,b)``, the latest bugbear release raises `zip() without an explicit strict= parameter.`
However, on python 3.7-3.9 this parameter is not present and therefore the check should ignore them

kasium:
Closes #320


README.rst
```diff
@@ -202,7 +202,7 @@ for details.
 
 **B905**: ``zip()`` without an explicit `strict=` parameter set. ``strict=True`` causes the resulting iterator
 to raise a ``ValueError`` if the arguments are exhausted at differing lengths. The ``strict=`` argument
-was added in Python 3.10, so don't enable this flag for code that should work on <3.10.
+was added in Python 3.10, so the check is disabled when running on python <3.10.
 For more information: https://peps.python.org/pep-0618/
```
jakkdl, line 206: I think this should still warn that you shouldn't enable this if you want your code to work on <3.10, even if flake8 is ran on python version 3.10+

cooperlees:
Yeah, I get the goal here but I'm still torn. Happy to go with it if most people think it's better behaviour.

README.rst
kasium: Any suggestions about the text?

Zac-HD (changes requested):
I think we should *not* do this: the check is already disabled by default, and if the user explicitly enables B905 I think it should be **possible** to use Python 3.9 to check code which requires Python 3.10.

kasium:
Thanks for the input. I guess then let's drop this

PR #328, opened 2023-01-09
Add B028 - suggest !r for quote-wrapped variables in f-strings

### LINKED ISSUE #319: Suggest `!r` format specifier for strings like `f"foo is '{self.foo}'"`
I've seen some code lately where a reminder about the `!r` format specifier would probably have helped - instead of displaying the repr, single-quotes were manually added outside the braces as in `f"No value found at '{path}'"`.  This OK, _until_ `path` contains a single-quote character and then it can get confusing fast.

I'd therefore propose warning about this case, which is conveniently easy to reliably detect [using the `JoinedStr` and `FormattedValue` nodes](https://docs.python.org/3/library/ast.html#ast.FormattedValue) - within a `JoinedStr`, warn if there is a sequence of `Constant`/`FormattedValue`/`Constant`, where the first constant ends and the second starts with a single or double quote character, and the formatter value has no conversion code.

Separately, I used the regex `".*('\{[^}]+\}').*"` to confirm that this is common enough to be worth linting for.  It's not especially rare, and IMO pretty useful to teach people about format specifiers.

jakkdl:
fixes #319 

I decided not to implement 3.7 support at all, expecting it to be dropped soon enough / be rare enough that it's not worth the added code complexity to support it.
3.8 is somewhat of a pain with not having `ast.unparse`, but this implementation should be decent-ish even when there's complex contents of an f-string variable.

This should likely not be merged until edge cases in #319 has been sorted out.

@Zac-HD

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -169,6 +169,8 @@ limitations make it difficult.
 
 **B027**: Empty method in abstract base class, but has no abstract decorator. Consider adding @abstractmethod.
 
+**B028**: Consider replacing ``f"'{foo}'"`` with ``f"{foo!r}"`` which is both easier to read and will escape quotes inside ``foo`` if that would appear. The check tries to filter out any format specs that are invalid together with ``!r``. If you're using other conversion flags then e.g. ``f"'{foo!a}'"`` can be replaced with ``f"{ascii(foo)!r}"``. Not currently implemented for python<3.8 or ``str.format()`` calls.
+
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


cooperlees (changes requested):
Sounds good to do >= 3.8 - Lets just state that in the README.rst description of this check please (that is missing here).

3.7 still has another ~6 months of life, we've kind of decided to follow python supported versions and follow flake8 version dropping. So I would expect 3.7 to hang around for little while longer ...

jakkdl:
Added Readme and Change Log entry.

Now won't warn if there's any format specifiers only valid for numerics - or width specified (which will misplace quotes relative to the padding).

And you said you wanted lots of tests, so there's now >90k checks :sweat_smile: The format spec permutation test takes a few seconds to run on my machine, but not much slower than the existing fuzz tests. I can scale it back if you think it's overkill though

cooperlees (approved):
Thanks John! All looks good to me. Will see if @Zac-HD pops past at all to double check :)

Zac-HD (approved):
Looks fantastic - thanks John!

Issue #278, opened 2022-08-24
False positive B024


kmike:
Hey! A new release (with https://github.com/PyCQA/flake8-bugbear/issues/273 included) raises an error

> Injectable is an abstract base class, but it has no abstract methods. Remember to use @ abstractmethod, @abstractclassmethod and/or @abstractproperty decorators.

I think that's fine to use ABC without abstractmethods/etc., because that's not the only thing ABCs do. They also allow to register other classes as being subclasses of the base class - see example at https://docs.python.org/3/library/abc.html#abc.ABCMeta.register. 

This is how ABC is used in our project (see https://github.com/scrapinghub/web-poet/blob/944dcd75f6c56988d42c56c56adb1b3dcafa2dbd/web_poet/pages.py#L11); a new flake8-bugbear release flags it as an error.

alk-gpajot:
Also, ABC could just be there to declare a class variable that should be set for concrete classes in a less verbose way than using annotations and properties:
```python
class Abstract(ABC):
    attr: ClassVar[str]
```
This is checked by mypy and will also raise an `AttributeError` if trying to access `attr` on a concrete class that doesn't set it. See https://peps.python.org/pep-0526/

LefterisJP:
> This is how ABC is used in our project (see https://github.com/scrapinghub/web-poet/blob/944dcd75f6c56988d42c56c56adb1b3dcafa2dbd/web_poet/pages.py#L11); a new flake8-bugbear release flags it as an error.

This is how it's used in our project too: https://github.com/rotki/rotki/blob/2c2bc209ba1d51663a86f4b710dfa6bbc24e7535/rotkehlchen/data_import/importers/binance.py#L43-L97

I don't think this is a good check to have and would opt to remove it.

mike0sv:
Also got false positive for this case:
```
class A(ABC):
   @abstactmethod
   def f1(self): pass

   @abstactmethod
   def f2(self): pass

class B(A, ABC):
   def f2(self): pass
```

kmike:
Hey! What's the reason the issue is closed? Is it done on purpose (essentially, "wontfix"), or is it a side-effect of some commit message?

cooperlees:
The comments above indicate a fix was landed (via #281). If you believe the latest version still does hi-lights this issue still or incorrectly still we could reopen it.

A repro, example code (if it differs in any way) or suggestions on how to make this better always welcome.

kmike:
[...]

PR #336, opened 2023-01-18
B024: don't warn on classes without methods

jakkdl:
Fix another case in #278 - so now it won't warn on empty classes or any class without methods, esp since the fix proposed in the error message (consider adding `@abstractmethod`) is not useful in that case.
@Zac-HD @mikhail-melnik

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,7 +154,9 @@ positives due to similarly named user-defined functions.
 the loop, because `late-binding closures are a classic gotcha
 <https://docs.python-guide.org/writing/gotchas/#late-binding-closures>`__.
 
-**B024**: Abstract base class with no abstract method. You might have forgotten to add @abstractmethod.
+**B024**: Abstract base class has methods, but none of them are abstract. This
+is not necessarily an error, but you might have forgotten to add the @abstractmethod
+decorator, potentially in conjunction with @classmethod, @property and/or @staticmethod.
 
 **B025**: ``try-except`` block with duplicate exceptions found.
 This check identifies exception types that are specified in multiple ``except``
@@ -311,6 +313,7 @@ Change Log
 Future
 ~~~~~~~~~
 
+* B024: now ignores classes without any methods.
 * B906: Ignore ``visit_`` functions with a ``_fields`` attribute that can't contain ast.AST subnodes. (#330)
 
 23.1.17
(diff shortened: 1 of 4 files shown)


jakkdl:
The error message in the readme and the message being raised were very different, so synced them to match better, and updated the message to be "softer".

bugbear.py
```diff
@@ -751,7 +753,7 @@ def is_str_or_ellipsis(node):
                     B027(stmt.lineno, stmt.col_offset, vars=(stmt.name,))
                 )
 
-        if not has_abstract_method:
+        if not has_abstract_method and has_method:
```
Zac-HD, line 756: tiny nit: this code is entirely correct, but writing `if has_method and not has_abstract_method:` makes it *obvious* that there's no operator precedence issue, and I find it better matches natural-language grammar.

Not asking for a change, just an FYI.

Zac-HD (approved):
Looks good to me, thanks @jakkdl!

cc @cooperlees in case you want to review, or plan a release?

bugbear.py
jakkdl: haha, that's a great nit. No reason not to change it~

README.rst
```diff
@@ -154,7 +154,9 @@ positives due to similarly named user-defined functions.
 the loop, because `late-binding closures are a classic gotcha
 <https://docs.python-guide.org/writing/gotchas/#late-binding-closures>`__.
 
-**B024**: Abstract base class with no abstract method. You might have forgotten to add @abstractmethod.
+**B024**: Abstract base class, with no abstract methods. This is not necessarily an
```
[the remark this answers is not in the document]
jakkdl: Sounds good.

And yeah, added to changelog.

jakkdl:
Fixed review comments and rebased on top of main

bugbear.py
```diff
@@ -1477,9 +1479,10 @@ def visit_Lambda(self, node):
 B023 = Error(message="B023 Function definition does not bind loop variable {!r}.")
 B024 = Error(
     message=(
-        "B024 {} is an abstract base class, but it has no abstract methods. Remember to"
-        " use the @abstractmethod decorator, potentially in conjunction with"
-        " @classmethod, @property and/or @staticmethod."
+        "B024 {} is an abstract base class, but none of them are abstract. This is "
```
Zac-HD, line 1482:
```suggestion
        "B024 {} is an abstract base class, but none of the methods it defines are abstract. This is "
```

?

jakkdl:
pre-commit.ci autofix

Issue #334, opened 2023-01-17
B017 support the matches keyword argument to pytest.raises


toofar:
#317 added support for warning about B017 when using `pytest.raises(Exception)`, and it included an exception for passing a regex to match against similar to unittest.assertRaisesRegex. But you can pass that regex as a keyword argument too and that exception doesn't apply in that case.

Docs with an example of passing the `match` keyword: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

Reproducer:

```python
import pytest


def test_positional_arg():
    with pytest.raises(Exception, 'some-regex'):  # this is fine
        pass


def test_keyword_arg():
    with pytest.raises(Exception, match='some-regex'):  # this gets a B017 warning
        pass
```

PR #337, opened 2023-01-18
B017: don't warn when ``pytest.raises()`` has a ``match`` argument

nsoranzo:
Fix https://github.com/PyCQA/flake8-bugbear/issues/334 .

Also add more tests for when we should and shouldn't emit B017.
In particular we should check ``assertRaises`` and ``pytest.raises`` calls outside of with statements.


nsoranzo:
> Can we add a 1 line into the Changes section of README.rst please.

Done in a rebased commit.

cooperlees (approved):

PR #339, opened 2023-01-23
Recommend `.removeprefix()` or `.removesuffix()` in `B005`

mxr:


cooperlees (approved):
Thanks. These are also good suggestions. I see no reason to not offer them to developers.

mxr:
Do you mean "no reason to not offer them"?

PR #340, opened 2023-01-29
Update isort pre-commit hook

FozzieHi:
This PR updates the isort pre-commit hook from 5.11.4 to 5.12.0.

On running the `pre-commit run --all-files` command for the first time, I encountered an error. This seemed to be the same error that was fixed in [this PR for isort](https://github.com/PyCQA/isort/pull/2078), and the update fixes the error.


cooperlees (approved):
Thanks!

Wonder if we have auto updates enabled here, if not we should. I'll need to remember how to check for that.

PR #341, opened 2023-01-29
B016: Warn when raising f-strings

FozzieHi:
Warns when raising f-strings which I think matches the check's current description.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -319,6 +319,7 @@ Future
   functions that are ignored by the B906 check. The ``ast.Bytes``, ``ast.Num`` and
   ``ast.Str`` nodes are all deprecated, but may still be used by some codebases in
   order to maintain backwards compatibility with Python 3.7.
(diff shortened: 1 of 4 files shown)


cooperlees (approved):
Thanks. Surprised this isn't a Python error, but neither are non f-strings so thanks for adding this! Wonder how many it will find around the world.

I wish we somehow had stats of all this plugin finds!

jakkdl:
> I wish we somehow had stats of all this plugin finds!

There's definitely ways of estimating those numbers, estimating the baseline prevalence of an error in projects without bugbear and multiplying by [estimated] number of flake8-bugbear installs. Or checking the number of hits on github search for the invalid pattern before and after implementing a new check. But that will underestimate the impact, as it only checks in pushed code and not thorny-bug-that-takes-much-longer-to-find-without-bugbear-but-will-get-fixed.

PR #342, opened 2023-01-30
Add a lint checking for warnings.warn stacklevel argument

### LINKED ISSUE #299: Require warnings.warn stacklevel argument as 2+
Calling [`warnings.warn()`](https://docs.python.org/3/library/warnings.html#warnings.warn) without a `stacklevel`, or a `stacklevel` value of <2, is nearly always a mistake. Doing so flags the warning inside the called function, rather than the caller, making it hard to fix the warning.

flake8-bugbear could detect such invocations of `warnings.warn()`.

This issue inspired by @asottile's video of the day: https://www.youtube.com/watch?v=CtFdXBEwYfk . This is also something I've seen recur as a review point in Django PR's.

FozzieHi:
This PR adds a lint which checks for an explicitly defined `stacklevel` keyword argument on the `warn` method from the `warnings` module. If `stacklevel` is not explicitly defined, this lint triggers. See #299 for more information.

I am new to this project and ASTs in general, so any feedback is appreciated! Specifically I don't know whether this should be `B028` or `B029`, as `B028` was used before by `B907`.

Closes #299

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1527,8 +1527,7 @@ B028 = Error(
         " warnings module uses a stacklevel of 1 by default. This will only show a"
         " stack trace for the line on which the warn method is called."
         " It is therefore recommended to use a stacklevel of 2 or"
-        " greater to provide more information to the user. This warning can be disabled"
-        " by explicitly setting a stacklevel."
+        " greater to provide more information to the user."
     )
 )
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -360,6 +360,7 @@ class BugBearVisitor(ast.NodeVisitor):
             self.check_for_b026(node)
 
         self.check_for_b905(node)
+        self.check_for_b028(node)
         self.generic_visit(node)
 
     def visit_Module(self, node):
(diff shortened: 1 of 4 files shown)


bugbear.py
```diff
@@ -1510,6 +1521,16 @@ def visit_Lambda(self, node):
...
     )
 )
+B028 = Error(
+    message=(
+        "B028 No explicit stacklevel keyword argument found. The warn method from the"
+        " warnings module uses a stacklevel of 1 by default. This will only show a"
+        " stack trace for the line on which the warn method is called."
+        " It is therefore recommended to use a stacklevel of 2 or"
+        " greater to provide more information to the user. This warning can be disabled"
+        " by explicitly setting a stacklevel."
```
cooperlees, line 1531:
```suggestion
        " greater to provide more information to the user."
```
I feel the opening sentence states this.

cooperlees (approved):
Many thanks. I've seen this a lot and is a good new addition to me. I'm also happy to reuse B028 since we renamed the old check to be optional ...

Will leave for 24 hours to let others comment.

bugbear.py
FozzieHi: Thanks! Updated now.

jakkdl:
Since B028 had a bunch of false positives I suspect there are going to be some configs out there that disabled it, and unless they closely track development haven't re-enabled it since it became optional. But I suspect there's not a ton of those, and B028 was renamed within three days, so it's probably fine to expect those people to update their configs. But one could also let B028 be for some time before using it again.

(it would be nice if there was a warning for disabled checks that don't exist... :thinking:)

PR #343, opened 2023-02-02
Fix BO28 README typo

FozzieHi:
Not entirely sure how this made it in without me noticing, are O and 0 really that similar in terms of font style? Sorry! :facepalm:


cooperlees (approved):
LoL - I definitely missed this ...

PR #344, opened 2023-02-02
Split B001 into B029 for empty tuple exception handling

### LINKED ISSUE #282: Confusing B001 message when catching empty tuple
This code
```python
    try:
        raise BaseException()
    except ():
        pass
```
gives this error message

> B001 Do not use `except ():`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.

However, `except ():` doesn't catch anything at all! Running the above code raises a `BaseException`. Perhaps a warning is warranted here, but with a different message.

FozzieHi:
This splits the part of B001's lint which checked for using `except` with an empty tuple, and moves it into a new lint, B029. As suggested by the comments in #282. Closes #282.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -177,7 +177,7 @@ limitations make it difficult.
 stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called.
 It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
 
-**B029**: Using ``except`` with an empty tuple does not handle anything. Add exceptions to handle instead.
+**B029**: Using ``except: ()`` with an empty tuple does not handle/catch anything. Add exceptions to handle.
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


bugbear.py
```diff
@@ -1530,6 +1527,12 @@ def visit_Lambda(self, node):
         " greater to provide more information to the user."
     )
 )
+B029 = Error(
+    message=(
+        "B029 Using `except` with an empty tuple does not handle anything. "
+        "Add exceptions to handle instead."
```
cooperlees, line 1533:
```suggestion
        "B029 Using `except: ()` with an empty tuple does not handle/catch "
        "anything. Add exceptions to handle."
```

Yup - that is also a '0' !

README.rst
```diff
@@ -325,6 +327,7 @@ Future
   order to maintain backwards compatibility with Python 3.7.
 * B016: Warn when raising f-strings.
 * Add B028: Check for an explicit stacklevel keyword argument on the warn method from the warnings module.
+* Add B029: Check when trying to use ``except`` with an empty tuple.
```
cooperlees, line 330:
```suggestion
* Add B029: Check when trying to use ``except`` with an empty tuple i.e. ``except: ()``.
```

cooperlees (approved):
LGTM - Thanks for digging though issues and just doing them! Legend.

FozzieHi:
Thanks! Changes look good to me.

PR #346, opened 2023-02-03
Introduce B030; fix crash on weird except handlers

### LINKED ISSUE #345: Crash on nested tuple in except handler
Run flake8-bugbear on this file:
```python
try: x
except (a, (b, c)): pass
```

And you will get a stack trace that ends with

```
  File "/Users/jelle/py/venvs/py311/lib/python3.11/site-packages/bugbear.py", line 283, in visit_ExceptHandler
    names = [_to_name_str(e) for e in node.type.elts]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jelle/py/venvs/py311/lib/python3.11/site-packages/bugbear.py", line 283, in <listcomp>
    names = [_to_name_str(e) for e in node.type.elts]
             ^^^^^^^^^^^^^^^
  File "/Users/jelle/py/venvs/py311/lib/python3.11/site-packages/bugbear.py", line 202, in _to_name_str
    return _to_name_str(node.value)
                        ^^^^^^^^^^
AttributeError: 'Tuple' object has no attribute 'value'
```

Why would you use nested tuples in except handlers, you may ask. I'm not sure, but it's legal syntax so I am testing it in the [pyanalyze](/quora/pyanalyze) test suite.

JelleZijlstra:
Fixes #345.

This introduces a new bugbear check against things like `except call():` or `except 1:`. This could be legitimate code (if you have a function that returns an exception class), but I'm not sure that's something that would ever come up in practice.

bugbear would previously crash on any except handler that is not a Tuple, Name, Attribute, or Call.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -177,6 +177,8 @@ limitations make it difficult.
 stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called.
 It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
 
+**B029**: Except handlers should only be exception classes or tuples of exception classes.
+
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -189,6 +189,50 @@ def _is_identifier(arg):
     return re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", arg.s) is not None
 
 
+def _flatten_excepthandler(node):
+    if isinstance(node, ast.Tuple):
+        for elt in node.elts:
+            yield from _flatten_excepthandler(elt)
+    else:
+        yield node
+
+
+def _check_redundant_excepthandlers(names, node):
+    # See if any of the given exception names could be removed, e.g. from:
+    #    (MyError, MyError)  # duplicate names
+    #    (MyError, BaseException)  # everything derives from the Base
+    #    (Exception, TypeError)  # builtins where one subclasses another
+    #    (IOError, OSError)  # IOError is an alias of OSError since Python3.3
+    # but note that other cases are impractical to handle from the AST.
+    # We expect this is mostly useful for users who do not have the
+    # builtin exception hierarchy memorised, and include a 'shadowed'
+    # subtype without realising that it's redundant.
+    good = sorted(set(names), key=names.index)
+    if "BaseException" in good:
+        good = ["BaseException"]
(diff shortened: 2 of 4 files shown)


cooperlees:
haha - I always feel bad when I push something somewhere and black is upset at my formatting! I can try fix before merge eventually too if you don't beat me.

cooperlees (approved):
LGTM - Thanks mate. One of us can just fix the formatting.

JelleZijlstra:
Looks like you just merged a different B029 so there's a little more work to do :)

I turned off my editor's format-on-save because it's annoying when writing test cases for Black, lol. Should probably figure out a way to keep it on for other projects.

JelleZijlstra:
pre-commit.ci autofix

JelleZijlstra:
All green now.

FozzieHi:
LGTM, but do we also want to allow function calls (by allowing ast.Call)?

For example:
```python
def exceptions():
    return ValueError, KeyError


try:
    pass
except exceptions():
    pass
```

Not entirely sure how common it is, and we obviously can't check the return of the function, but I think it'd still be useful for some use cases.

JelleZijlstra:
I guess we could allow it. People could also plausibly write `except ValueError(): pass` instead of `except ValueError: pass`, which this check would help catch.

@cooperlees what do you think, should we allow calls in except handlers?

FozzieHi:
That's a good point. I think we definitely want to catch that `except ValueError()` case, and I think anyone using the uncommon function call can simply just `noqa` it. I'm fine with keeping it how it is.

cooperlees:
This is a strange pattern, but I am sure valid. Is there any harm doing a function call here? I don't see any so probably best we handle this edge case ... But happy for another PR to do that ... Was going to release a new version later today.

FozzieHi:
What Jelle mentioned above, `except ValueError():` results in a TypeError. In this current version we would catch that. If we added `ast.Call` to the permitted values then we wouldn't catch that, but we would allow for this, which is valid:
```python
def exceptions():
    return ValueError, KeyError


try:
    pass
except exceptions():
    pass
```

I think that we would want to catch the `except ValueError():` over allowing the above valid code (so we shouldn't add `ast.Call` to the list of permitted nodes). With the user being able to `noqa` any valid code.

cooperlees:
Yeah, lets see if it's a common pattern (via issues opened) and if so address it then. Thanks again all!

PR #354, opened 2023-02-15
B030: allow splats and calls

### LINKED ISSUE #352: B030 doesn't allow splatted exceptions
I *think* there's a false-report coming from B030 where we have some exceptions types defined in a constant. A minimal example:

```python
EXCEPTIONS_TO_IGNORE = (
    ValueError,
    TypeError,
)

try:
    pass
except (SyntaxError, *EXCEPTIONS_TO_IGNORE):
    pass
```

```console
$ flake8 t.py
t.py:8:1: B030 Except handlers should only be names of exception classes
```

This isn't incorrect, so I don't *think* the intention of B030 is to stop this kind of thing. 

<details>
<summary>versions etc</summary>

I'm using Macos 13.1, we're also seeing this in CI using Ubuntu 20.04.5.

```console
$ python --version --version
Python 3.10.8 (main, Oct 13 2022, 10:17:43) [Clang 14.0.0 (clang-1400.0.29.102)]

$ flake8 --version
6.0.0 (flake8-bugbear: 23.2.13, mccabe: 0.7.0, pycodestyle: 2.10.0, pyflakes: 3.0.1) CPython 3.10.8 on Darwin
```

</details>

JelleZijlstra:
Fixes #352.

I allow calls in addition to splats because a real-world false positive was reported
and buggy code doesn't seem common enough to merit an on-by-default error.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -325,6 +325,11 @@ MIT
 Change Log
 ----------
 
+Unreleased
+~~~~~~~~~~
+
+* B030: Allow calls and starred expressions in except handlers.
+
 23.2.13
 ~~~~~~~~~
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -333,17 +333,25 @@ class BugBearVisitor(ast.NodeVisitor):
         handlers = _flatten_excepthandler(node.type)
         good_handlers = []
         bad_handlers = []
+        ignored_handlers = []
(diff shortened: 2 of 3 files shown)


cooperlees (approved):
Thanks Jelle. Makes sense.

Issue #162, opened 2021-03-25
Ignore # type: ignore  and # noqa for line length


Dreamsorcerer:
When calculating the line length for B950, it would be nice to ignore trailing `# type: ignore` and `# noqa` comments.

These comments are typically not important and are just working around some issues with static checkers, so it tends to be bothersome when adding them results in more violations.

Formats that should be ignored for line length:
```
# type: ignore
# type: ignore[some-code]
# noqa
# noqa: B100
# type: ignore[some-code] # noqa: B100
```

cooperlees:
Sure, sounds good to me. We already upset PEP8 purists, so I think this is also fine. Would live another perspective tho - Maybe @ambv ?

Happy to accept a PR that adds this and shows it working with unittests! Thanks for the idea.

dejvidq:
I created PR #358 with my proposed solution.

Feel free to check.

FozzieHi:
@cooperlees This can be closed now :)

Dreamsorcerer:
#358

PR #358, opened 2023-02-16
B950: Ignores 'noqa' and 'type: ignore' comments (#162)

dejvidq:
Implements #162 

`noqa` and `type: ignore` comments won't be taken into account when checking lines.

In my proposal I use a regular expression, to remove and thus ignore the comments mentioned in the issue.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -68,28 +68,29 @@ class BugBearChecker:
 
         The following simple checks are based on the raw lines, not the AST.
         """
-        noqa_type_ignore_regex = re.compile(r"\s*#\s*(noqa:|type:\s*ignore).*$")
+        noqa_type_ignore_regex = re.compile(r"#\s*(noqa|type:\s*ignore)[^#\r\n]*$")
         for lineno, line in enumerate(self.lines, start=1):
             # Special case: ignore long shebang (following pycodestyle).
             if lineno == 1 and line.startswith("#!"):
                 continue
 
             # At first, removing noqa and type: ignore trailing comments"
-            line = noqa_type_ignore_regex.sub("", line)
+            line_without_first_comment = noqa_type_ignore_regex.sub("", line)
+            no_comment_line = noqa_type_ignore_regex.sub("", line_without_first_comment)
 
-            length = len(line) - 1
-            if length > 1.1 * self.max_line_length and line.strip():
+            length = len(no_comment_line) - 1
+            if length > 1.1 * self.max_line_length and no_comment_line.strip():
                 # Special case long URLS and paths to follow pycodestyle.
                 # Would use the `pycodestyle.maximum_line_length` directly but
                 # need to supply it arguments that are not available so chose
                 # to replicate instead.
-                chunks = line.split()
+                chunks = no_comment_line.split()
 
                 is_line_comment_url_path = len(chunks) == 2 and chunks[0] == "#"
 
                 just_long_url_path = len(chunks) == 1
 
-                num_leading_whitespaces = len(line) - len(chunks[-1])
+                num_leading_whitespaces = len(no_comment_line) - len(chunks[-1])
                 too_many_leading_white_spaces = (
                     num_leading_whitespaces >= self.max_line_length - 7
                 )
--- a/bugbear.py
+++ b/bugbear.py
@@ -75,8 +75,9 @@ class BugBearChecker:
                 continue
 
             # At first, removing noqa and type: ignore trailing comments"
-            line_without_first_comment = noqa_type_ignore_regex.sub("", line)
-            no_comment_line = noqa_type_ignore_regex.sub("", line_without_first_comment)
+            no_comment_line = noqa_type_ignore_regex.sub("", line)
+            if no_comment_line != line:
+                no_comment_line = noqa_type_ignore_regex.sub("", no_comment_line)
 
             length = len(no_comment_line) - 1
             if length > 1.1 * self.max_line_length and no_comment_line.strip():
--- a/bugbear.py
+++ b/bugbear.py
@@ -68,10 +68,14 @@ class BugBearChecker:
 
         The following simple checks are based on the raw lines, not the AST.
         """
+        noqa_type_ignore_regex = re.compile(r"\s*#\s*(noqa:|type:\s*ignore).*$")
         for lineno, line in enumerate(self.lines, start=1):
             # Special case: ignore long shebang (following pycodestyle).
             if lineno == 1 and line.startswith("#!"):
                 continue
+
+            # At first, removing noqa and type: ignore trailing comments"
+            line = noqa_type_ignore_regex.sub("", line)
 
             length = len(line) - 1
             if length > 1.1 * self.max_line_length and line.strip():
--- a/tests/b950.py
+++ b/tests/b950.py
@@ -24,9 +24,13 @@
 """
                                                                                                                                                                 
 """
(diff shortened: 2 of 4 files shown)


tests/b950.py
```diff
@@ -24,3 +24,9 @@
 """
                                                                                                                                                                 
 """
+"https://foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com  # noqa: F401"
+"https://foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com#noqa:F401, B950"
+"https://foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com  # type: ignore[some-code]"
+"https://foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com # type: ignore[some-code]"
+"https://foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com # type: ignore[some-code] # noqa: F401"
+"https://foooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com  # noqa: F401 # type:ignore[some-code]"
```
Dreamsorcerer, line 32: I'd suggest a couple of checks without codes, as per the examples in my issue (`# noqa` and `# type: ignore`.

Pretty sure the code works correctly, but best to cover them in the tests.

Dreamsorcerer (approved):

FozzieHi:
Thanks for this. Few things I've noticed:

1) This regex is vulnerable to ReDoS using the built-in `re` engine. Take this as an example, which will take longer as you increase the `\r` count:
```python
import re

noqa_type_ignore_regex = re.compile(r"\s*#\s*(noqa:|type:\s*ignore).*$")

print(noqa_type_ignore_regex.sub("", '\r' * 50_000 + 'noqa:'))
```
2) It doesn't match blanket `noqa` directives (e.g. `# noqa`).

Dreamsorcerer:
>     2. It doesn't match blanket `noqa` directives (e.g. `# noqa`).

Ah, I missed the colon, you're right.

I think this would probably work, and doesn't seem to be an issue with the above example (matching the preceding whitespace seems redundant).

`noqa_type_ignore_regex = re.compile(r"#\s*(noqa|type\s*ignore).*$")`


It does also have false positives though, like:
`skip_comment = "# noqa"`

Realistically, it probably won't cause any issues, but there was a better way to detect comments, it would be better.

FozzieHi:
@Dreamsorcerer Thanks. That's still vulnerable to ReDoS with the `re` engine, for example:
```python
import re

noqa_type_ignore_regex = re.compile(r"#\s*(noqa|type\s*ignore).*$")

print(noqa_type_ignore_regex.sub("", '#noqa' * 20_000 + '\nno_match'))
```

I think if we're going to continue with using regex to match this, switching to a linear time complexity regex engine would be a good move.

bugbear.py
```diff
@@ -68,11 +68,15 @@ def gen_line_based_checks(self):
...
+        noqa_type_ignore_regex = re.compile(r"\s*#\s*(noqa:|type:\s*ignore).*$")
         for lineno, line in enumerate(self.lines, start=1):
             # Special case: ignore long shebang (following pycodestyle).
             if lineno == 1 and line.startswith("#!"):
                 continue
 
+            # At first, removing noqa and type: ignore trailing comments"
+            line = noqa_type_ignore_regex.sub("", line)
+
             length = len(line) - 1
```
cooperlees, line 80:
```suggestion
            no_comment_line = noqa_type_ignore_regex.sub("", line)

            length = len(no_comment_line) - 1
```
Let's maybe be more explicit in variable name to make it clearer to people in the future reading this ...

cooperlees (commented):
Looks good to me, there is no nice way to parse comments to my knowledge, but I generally love to avoid regexes wherever possible but I don't have a good suggestion here.

I also don't want to move this project from a pure python project if we can avoid it unless flake8 does. I don't want to get issues about compiling issues and increase potential testing and deployment issues.

tests/b950.py
cooperlees: Lets please cover all we exclude please for completeness

FozzieHi:
@cooperlees I agree, and I personally don't want to add dependencies if it would cause issues. However, given that it's feasible that linters are ran on untrusted input I don't think we should be adding any regexes which could potentially cause catastrophic backtracking (even though the risk when running against legitimate input seems to be very low).

If we want to keep using `re`, we just need to be careful what we match at the end of the regex (i.e. we don't want to match `#` as this can cause catastrophic backtracking).

For example, this regex is safe: `#\s*(noqa|type:\s*ignore)[^#\r\n]*$`. This will get rid of most of what we want, however it won't remove whitespace before the start of the comment. It also won't remove all of the comments within multiple ignore comments, e.g. `# noqa: F401 # type:ignore[some-code]`. Here it **will** remove the ending comment (`# type:ignore[some-code]`) but not the comment before that.

cooperlees:
> Maybe do 2 regex passes? 1 for noqa and 1 for type:ignore?

If this makes it safer and avoids a C extension, I am all for it. If we ever get to slow, we can reassess then.

FozzieHi:
Running a second regex pass if the first changes anything seems to be a good solution and it seems like it would catch most ignore comments.

cooperlees (changes requested):
We seem to have unittests failing ...

tests/test_bugbear.py
```diff
@@ -689,6 +689,8 @@ def test_b950(self):
                 B950(12, 103, vars=(103, 79)),
                 B950(14, 103, vars=(103, 79)),
                 B950(21, 97, vars=(97, 79)),
+                B950(35, 104, vars=(104, 79)),
+                B950(36, 104, vars=(104, 79)),
```
Dreamsorcerer, line 693: Guessing this is meant to be copied to the next 2 tests as well.

bugbear.py
```diff
@@ -68,24 +68,29 @@ def gen_line_based_checks(self):
...
         for lineno, line in enumerate(self.lines, start=1):
             # Special case: ignore long shebang (following pycodestyle).
             if lineno == 1 and line.startswith("#!"):
                 continue
 
-            length = len(line) - 1
-            if length > 1.1 * self.max_line_length and line.strip():
+            # At first, removing noqa and type: ignore trailing comments"
+            line_without_first_comment = noqa_type_ignore_regex.sub("", line)
+            no_comment_line = noqa_type_ignore_regex.sub("", line_without_first_comment)
```
FozzieHi, line 79:
```suggestion
            no_comment_line = noqa_type_ignore_regex.sub("", line)
            if no_comment_line != line:
                no_comment_line = noqa_type_ignore_regex.sub("", no_comment_line)
```

We only need to run a second regex pass if the first one actually changes anything.

FozzieHi (changes requested):
LGTM apart from the tests and a minor comment regarding the second regex pass.

dejvidq:
> We seem to have unittests failing ...

Yes, I'm sorry, my bad. After the refactor and adding new tests, I forgot to run whole tests suite and ran only `test_b950` 😐

bugbear.py
dejvidq: Changed 👍

cooperlees (approved):
Many thanks for sticking with it and getting it over the line!

Dreamsorcerer:
Could be nice to exclude `# pragma: no cover` as well.

PR #361, opened 2023-02-25
Fix typos in docstring

jaylangford:
Change the docstring to the correct code. Change the line numbers that are expected to trigger the warning code.

### DIFF
--- a/tests/b031.py
+++ b/tests/b031.py
@@ -1,6 +1,6 @@
 """
 Should emit:
-B030 - on lines 29, 33, 43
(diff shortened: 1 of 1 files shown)


jaylangford:
I wasn't 100% sure about the line numbers. But the first and third line numbers are inconsistent in their expectations. The new line numbers are what were actually printed by flake8 when I ran it against the file.

cooperlees (approved):
Thanks for this, this seems correct to me. I bet black reformatted a new line in or something and the author missed it.

PR #364, opened 2023-03-08
Fix bug in b030

aacunningham:
I was working in ruff implementing B030 when I noticed that the test case for the warning didn't look right. It may be that another rule catches this issue, but it made sense to me that B030 would catch it as well.

```python
try:
    raise ValueError()
except (ValueError, (RuntimeError, (KeyError, TypeError))):
    pass
```

This code causes an exception, `TypeError: catching classes that do not inherit from BaseException is not allowed`, because the inner tuple isn't expanded out automatically. It passes, however, if those tuples are starred:

```python
try:
    raise ValueError()
except (ValueError, *(RuntimeError, *(KeyError, TypeError))):
    pass
```

The wording for B030 makes it sound like this is the exact error it's meant to catch, so I updated the `_flatten_excepthandler` method to be more selective when flattening tuples and arrays. In the updated code, the `(RuntimeError, ...)` tuple is returned fully, which then gets it placed in the bad_handlers list.

Let me know if there's anything I missed, any improvements I can make, or if I should have started this convo in an issue. I had already gotten this logic working in Rust, so I tried to port it into python and plug it in to see if I could get the same result. Work was already done, so I figured a PR was a decent way to present it.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -190,11 +190,17 @@ def _is_identifier(arg):
 
 
 def _flatten_excepthandler(node):
-    if isinstance(node, ast.Tuple):
-        for elt in node.elts:
-            yield from _flatten_excepthandler(elt)
-    else:
+    if not isinstance(node, ast.Tuple):
         yield node
+        return
+    exprs_to_flatten = [node.elts]
+    while len(exprs_to_flatten):
+        exprs = exprs_to_flatten.pop()
+        for expr in exprs:
+            if isinstance(expr, ast.Starred) and isinstance(expr.value, (ast.List, ast.Tuple)):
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):
I actually started writing this in order to *allow* nested tuples, because I somehow convinced myself that it worked, but you're right that it actually doesn't work. Not sure what I was thinking.

aacunningham:
Also, it was getting late here when I wrote this up, but just a quick thanks for the project in the first place. Really appreciate y'all's work, and it's been a nice hands on introduction to the ast module (the one my coworkers import from instead of typing ;) )

cooperlees (approved):

Issue #365, opened 2023-03-09
Move to >= 3.8.1 like flake8 + maybe add a CI check somehow


cooperlees:
Bonus: Maybe add some CI to check if our python-requires matches flake8

We could request that latest simple page for each project (i.e. https://pypi.org/simple/flake8/ and https://pypi.org/simple/flake8-bugbear/) and compare the latest version's `data-requires-python`
- There is even a JSON version of these pages now ...

I think weekly would be enough to run the check

FozzieHi:
Are there any major blockers to this update?

There's some minor ones which ideally we'd want to change, for example there's quite a few deprecated ast methods since 3.8 such as `visit_Num()`, `visit_Str()`, `visit_Bytes()`, `visit_NameConstant()`, `visit_Ellipsis()` having all been replaced with `visit_Constant()`.

cooperlees:
> Are there any major blockers to this update?

Na, just need to start doing it and use pyup and move tooling to be >= 3.8 so we can move the code base forward.

> There's some minor things that ideally we'd want to change. For example, there's quite a few deprecated ast methods since 3.8 such as `visit_Num()`, `visit_Str()`, `visit_Bytes()`, `visit_NameConstant()`, `visit_Ellipsis()` having all been replaced with `visit_Constant()`.

Yup, we should take advantage of all the AST improvements. Any TODO's or comments that state we can update things too ...

All PRs welcome, small or large to get us there.

FozzieHi:
Yup, happy to submit a PR with what I mentioned above. Do you want to create a `3.8` branch or something similar, or do you want this straight into `main`?

cooperlees:
> Yup, happy to submit a PR with what I mentioned above. Do you want to create a `3.8` branch or something similar, or do you want this straight into `main`?

Happy to go to main and make next release >=3.8 ...

FozzieHi:
@cooperlees I think we've pretty much covered everything? We might want to change the `flake8` dependency requirement from `>=3.0.0` to `>=6.0.0` like I briefly mentioned in #368.

Not sure how a CI check would work, or if it's really needed given how infrequently the versions change.

cooperlees:
Yeah, I agree, the CI can wait, but I want to check if we match base flake8 better in the long run.

I agree with the >=6.0.0 flake8. Will try get to it or PR welcome and I'll get a release out as soon as we are as we also have a bug fix we need to get out ...

I might make a periodic action one day to check for this ...

PR #366, opened 2023-03-14
(3.8) Replace deprecated ast literals

FozzieHi:
Tracking issue: #365

See python/cpython#77073:
> Num -- for int, float and complex
Str -- for str
Bytes -- for bytes
Ellipsis -- for Ellipsis
NameConstant -- for True, False and None

This looks a bit messy in places but I don't think a helper function would provide that much improved readability (it'd effectively just be re-implementing the deprecation's, which we don't want), and we only use this in a relatively small portion of the code base.

I've just replaced the deprecated literals with their actual meanings, I'm not too sure whether we now want to change some (for example ast.Num checks for int, float and complex but we might only want to check for int).

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -189,7 +189,7 @@ def _is_identifier(arg):
     # Return True if arg is a valid identifier, per
     # https://docs.python.org/2/reference/lexical_analysis.html#identifiers
 
-    if not isinstance(arg, ast.Str):
+    if not isinstance(arg, ast.Constant) or not isinstance(arg.value, str):
         return False
 
     return re.match(r"^[A-Za-z_][A-Za-z0-9_]*$", arg.s) is not None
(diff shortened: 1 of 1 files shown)


FozzieHi:
Oh right, we might want to disable the 3.7 ci run first :facepalm:

cooperlees:
> Oh right, we might want to disable the 3.7 CI run first facepalm

Maybe we should do this in a PR we land before this turning off CI, adjusting pyproject.toml and anything else that ups our minimums for tools ...

FozzieHi:
@cooperlees Just read my mind, submitting a PR for that now.

cooperlees:
> @cooperlees Just read my mind, submitting a PR for that now.

I'm around for a few hours so can merge quickly to make rebasing this easier.

cooperlees (approved):
All tests pass and we have good coverage so pretty confident here.

PR #367, opened 2023-03-14
(3.8) Update CI and metadata

FozzieHi:
Tracking issue: #365

Updates CI to remove 3.7 runs and updates project metadata from 3.7 to 3.8.

Will be needed for CI to run on #366

PR #368, opened 2023-03-14
(3.8) Minor tweaks

FozzieHi:
Tracking issue: #365

Some minor changes:
- Changes in `bugbear.py` and `test_bugbear.py` from running `pyupgrade --py38-plus` against them (I left the individual test files, not sure if we want to change those)
- A change to `pyproject.toml` setting `requires-python` to `>=3.8.1` instead of `>=3.8` to match flake8
- A change I missed in `tox.ini` on #367

Should we also change the `flake8` dependency requirement from `>=3.0.0` to `>=6.0.0`?

### DIFF
--- a/tests/test_bugbear.py
+++ b/tests/test_bugbear.py
@@ -747,8 +747,7 @@ class BugbearTestCase(unittest.TestCase):
         filename = Path(__file__).absolute().parent.parent / "bugbear.py"
         proc = subprocess.run(
             ["flake8", str(filename)],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            capture_output=True,
             timeout=60,
         )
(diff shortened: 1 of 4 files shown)


tests/test_bugbear.py
```diff
@@ -747,8 +747,7 @@ def test_selfclean_bugbear(self):
         filename = Path(__file__).absolute().parent.parent / "bugbear.py"
         proc = subprocess.run(
             ["flake8", str(filename)],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
+            capture_output=True,
```
cooperlees, line 750: O, TIL, have never used ! NIce.

cooperlees (approved):
Nice - Agree with 3.8.1 - I remember this!

Thanks.

Issue #371, opened 2023-03-21
New check: duplicate types in type hint, or duplicate literals in set


jakkdl:
```python
a: int | int | float = ...
b: Union[List[int, int]] = ...
c: Optional[int] | int | None = ... # can probably rely on other linters to not have to deal with this one
d: set[int] = {1, 2, 3, 3, 5}

e: Tuple[int, int] = ... # not an error
```
these should pretty much always be typos and/or bad copy-pastes, with very few to no false alarms. Don't think it's super common, but definitely not rare or super esoteric.

The check shouldn't have to be too complicated to implement, for the type hint probably need a whitelist of what types you look in (List/list, Set/set, Dict/dict (with special logic), OrderedDict, Optional, Union, Literal...).

For sets you only need to look inside `set()` (ast.Call) and inside `{}` (ast.Set), and probably only include literal int/float/strings.

cooperlees:
This seems super uncontroversial if we can AST parse this out nicely covering the edge cases you've outlined and we can improve on reports ... I'd be happy to add this.

We can run this over some large type checked repos before releasing too to be extra safe.

AlexWaygood:
FYI we already have a check for duplicate items in a union over at [flake8-pyi](https://github.com/PyCQA/flake8-pyi) (error code Y016). Feel free to take inspiration from our implementation :)

FozzieHi:
I think these should be two different checks, but I'm happy to work on a duplicate items in a set check.

> For sets you only need to look inside set() (ast.Call) and inside {} (ast.Set), and probably only include literal int/float/strings.

What do you mean by checking `ast.Call`? Do we also want to catch these?
`set((1, 2, 3, 3, 5))`
`set([1, 2, 3, 3, 5])`

I'm not entirely sure how common these would be, and it would increase the complexity of the check, but I could be missing another syntax here!

jakkdl:
> I think these should be two different checks, but I'm happy to work on a duplicate items in a set check.
> 
> > For sets you only need to look inside set() (ast.Call) and inside {} (ast.Set), and probably only include literal int/float/strings.
> 
> What do you mean by checking `ast.Call`? Do we also want to catch these?
> `set((1, 2, 3, 3, 5))`
> `set([1, 2, 3, 3, 5])`
> 
> I'm not entirely sure how common these would be, and it would increase the complexity of the check, but I could be missing another syntax here!
> 

[...]

PR #373, opened 2023-03-22
Add B033: Duplicate items in sets

FozzieHi:
This checks for duplicate items in sets when using the `{}` syntax.

Adds a part of #371.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -310,7 +310,6 @@ class BugBearVisitor(ast.NodeVisitor):
     NODE_WINDOW_SIZE = 4
     _b023_seen = attr.ib(factory=set, init=False)
     _b005_imports = attr.ib(factory=set, init=False)
-    _b033_seen = attr.ib(factory=set, init=False)
 
     if False:
         # Useful for tracing what the hell is going on.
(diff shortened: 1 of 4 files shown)


FozzieHi:
Love it when I find a simpler solution within seconds of pushing. I'll blame it on me only using Python regularly for a few months!

bugbear.py
```diff
@@ -518,6 +518,10 @@ def visit_Import(self, node):
         self.check_for_b005(node)
         self.generic_visit(node)
 
+    def visit_Set(self, node):
```
JelleZijlstra, line 521: Can we do this for dict keys too? It should be fairly easy and it's also a reasonably common bug.

FozzieHi: Pyflakes (which is included in flake8) has a rule to detect duplicate keys with **different** values, [F601 dictionary key name repeated with different values](https://flake8.pycqa.org/en/latest/user/error-codes.html).

It doesn't flag duplicate keys with the same values, but as that's unlikely to cause a bug per se, should we check for that?

JelleZijlstra: Thanks for the extra context! Yes I think dicts with both keys and values the same are pretty much equivalent to duplicate set members. Honestly I don't have a great sense of what rules belong where, but maybe to avoid duplication we should only alert in cases where pyflakes doesn't?

FozzieHi: I agree, we would only want to flag when both the dictionary keys and values are the same to avoid conflicts with Pyflakes. 

I think a new rule would be good for this (in case people want to disable each rule separately) and we could leave B033 as it is? As we'd be checking both the keys and the values it also wouldn't easily fit into this rule anyway, so I don't think we'd gain much by combining them into a single rule.

tests/b033.py
```diff
@@ -0,0 +1,17 @@
...
+B033 - on lines 6-12
+"""
+
+test = {1, 2, 3, 3, 5}
+test = {"a", "b", "c", "c", "e"}
+test = {True, False, True}
+test = {None, True, None}
+test = {3, 3.0}
+test = {1, True}
+test = {0, False}
```
cooperlees, line 12: Do True and False booleans hash to the same as 0 and !0 ... TIL if so. Guess I've never thought about it.

FozzieHi: Yup, I also discovered this when creating the check.

```python
test = {1, True}
print(test)
```
Will print:
```
{1}
```

cooperlees (approved):

PR #375, opened 2023-03-27
(3.8) Remove outdated sys.version_info checks

FozzieHi:
Tracking issue: #365

Removes outdated `sys.version_info` checks which we don't need with our minimum Python version of 3.8.1.


cooperlees (approved):
Many thanks for the clean up! I missed this.

Issue #381, opened 2023-04-26
assertRaises-type tests should only contain a single statement


JelleZijlstra:
Consider this code:
```python
with pytest.raises(TypeError):
    1 + "x"
    "x" + 1
```

This will pass if only one of the two statements raises, which is probably not what the author wanted. Instead, each statement should be in its own `pytest.raises` context. (Real-world example: python/cpython#103880.)

We should show an error if an assertRaises-type context (`pytest.raises`, `self.assertRaises`, `self.assertRaisesRegex`) contains more than one top-level statement.

cooperlees:
Yup - sure, This is definitely a bug that could hide failures. PR would be accepted.

I'd like to keep the scope to testing here tho to start with. This could open a can of worms if we look at context mangers/exception handling more broadly ...

NewGlad:
Can be closed @cooperlees

PR #382, opened 2023-05-07
Add B908: Rule to check that assertRaises-like contexts has no more than 1 top-level statements

NewGlad:
Implementation of the proposal from https://github.com/PyCQA/flake8-bugbear/issues/381


bugbear.py
```diff
@@ -1104,6 +1105,29 @@ def check_for_b022(self, node):
         ):
             self.errors.append(B022(node.lineno, node.col_offset))
 
+    def check_for_b908(self, node: ast.With):
+        has_multiline_body = len(node.body) > 1
+        for node_item in node.items:
+            if (
+                not hasattr(node_item.context_expr, "func")
```
NewGlad, line 1112: It looks so ugly, but I did not find a better solution
Do you have any suggestions to improve this block, let's say use some type check?

JelleZijlstra: First I would make a helper function like `is_assertRaises_like`, so you can make it `return False/True` with slightly simpler conditionals. Then the body of this can just be `if is_assertRaises_like(node_item.context_expr) and has_multiline_body:`. And instead of the hasattr checks, you can do `isinstance` checks on the appropriate AST classes, like `isinstance(expr, ast.Call)`.

bugbear.py
```diff
@@ -1104,6 +1105,29 @@ def check_for_b022(self, node):
...
+                or not hasattr(node_item.context_expr.func, "value")
+                or not hasattr(node_item.context_expr.func.value, "id")
+            ):
+                continue
+            if (
```
JelleZijlstra, line 1117: I think we should also catch plain `with raises():`: https://grep.app/search?q=from%20pytest%20import%20raises

Also, let's cover `pytest.warns` too, it's the same idea.

bugbear.py
```diff
@@ -1104,6 +1105,29 @@ def check_for_b022(self, node):
...
+                # pytest form
+                node_item.context_expr.func.value.id == "pytest"
+                and node_item.context_expr.func.attr == "raises"
+            ) or (
+                # unittest form
+                node_item.context_expr.func.value.id == "self"
+                # to catch all variants like
+                # 'assertRaises', 'assertRaisesRegex', 'assertRaisesRegexp'
+                and node_item.context_expr.func.attr.startswith("assert")
```
JelleZijlstra, line 1126: I'd prefer to list all the names explicitly, since it's plausible people will have other `with self.assert...` like methods that don't work like the unittest ones. I think the list you have is good, but I'd also add `assertWarns` and its two regex forms.

tests/b908.py
```diff
@@ -0,0 +1,28 @@
+import unittest
```
JelleZijlstra, line 1: Should also add some tests with only one statement inside to make sure it does not trigger for those.

bugbear.py
```diff
@@ -1759,7 +1783,12 @@ def visit_Lambda(self, node):
         " flag."
     )
 )
-
+B908 = Error(
+    message=(
```
JelleZijlstra, line 1787: We should also describe the new warning in https://github.com/pycqa/flake8-bugbear#list-of-warnings. I think to many people it may not be obvious why this is a problem, so we should have a clear explanation.

JelleZijlstra (commented):
Thank you for your contribution!

bugbear.py
NewGlad: Thanks for the suggestion!

README.rst
```diff
@@ -228,6 +228,8 @@ This is meant to be enabled by developers writing visitors using the ``ast`` mod
 
 **B907**: Consider replacing ``f"'{foo}'"`` with ``f"{foo!r}"`` which is both easier to read and will escape quotes inside ``foo`` if that would appear. The check tries to filter out any format specs that are invalid together with ``!r``. If you're using other conversion flags then e.g. ``f"'{foo!a}'"`` can be replaced with ``f"{ascii(foo)!r}"``. Not currently implemented for python<3.8 or ``str.format()`` calls.
 
+**B908**: Contexts with exceptions assertions like ``with self.assertRaises`` or ``with pytest.raises`` should have not multiple top-level statements. Each statement should be in its own context.
```
JelleZijlstra, line 231:
```suggestion
**B908**: Contexts with exceptions assertions like ``with self.assertRaises`` or ``with pytest.raises`` should have not multiple top-level statements. Each statement should be in its own context. That way, the test ensures that the exception is raised only in the exact statement where you expect it.
```

bugbear.py
```diff
@@ -28,6 +28,14 @@
     ast.GeneratorExp,
 )
 FUNCTION_NODES = (ast.AsyncFunctionDef, ast.FunctionDef, ast.Lambda)
+B908_pytest_functions = {"raises", "warns"}
+B908_unittest_methods = {
+    "assertRaises",
+    "assertRaisesRegex",
+    "assertRaisesRegexp",
+    "assertWarns",
+    "assertWarnsRegex",
```
JelleZijlstra, line 37:
```suggestion
    "assertWarnsRegex",
    "assertWarnsRegexp",
```
Why not

bugbear.py
```diff
@@ -1104,6 +1113,38 @@ def check_for_b022(self, node):
...
+                # "with pytest.raises"
+                node.context_expr.func.value.id == "pytest"
+                and node.context_expr.func.attr in B908_pytest_functions
+            ) or (
+                # "with self.assertRaises"
+                node.context_expr.func.value.id == "self"
+                # to catch all variants like
+                and node.context_expr.func.attr in B908_unittest_methods
+            )
+
```
JelleZijlstra, line 1140: Missing an "else: return False"

bugbear.py
```diff
@@ -1104,6 +1113,38 @@ def check_for_b022(self, node):
...
+    def check_for_b908(self, node: ast.With):
+        if not len(node.body) > 1:
```
JelleZijlstra, line 1142:
```suggestion
        if len(node.body) == 1:
```

bugbear.py
NewGlad: There is no such method as I can see https://grep.app/search?q=assertWarnsRegexp
My python3.11 installation has no such method in `unittest`

bugbear.py
NewGlad: I prefer to have `if len(node.body) < 2` then

bugbear.py
NewGlad: but I can add it just in case

README.rst
```diff
@@ -228,6 +228,8 @@ This is meant to be enabled by developers writing visitors using the ``ast`` mod
 
 **B907**: Consider replacing ``f"'{foo}'"`` with ``f"{foo!r}"`` which is both easier to read and will escape quotes inside ``foo`` if that would appear. The check tries to filter out any format specs that are invalid together with ``!r``. If you're using other conversion flags then e.g. ``f"'{foo!a}'"`` can be replaced with ``f"{ascii(foo)!r}"``. Not currently implemented for python<3.8 or ``str.format()`` calls.
 
+**B908**: Contexts with exceptions assertions like ``with self.assertRaises`` or ``with pytest.raises`` should have not multiple top-level statements. Each statement should be in its own context. That way, the test ensures that the exception is raised only in the exact statement where you expect it.
+
```
cooperlees, line 232:
```suggestion
**B908**: Contexts with exceptions assertions like ``with self.assertRaises`` or ``with pytest.raises`` should not have multiple top-level statements. Each statement should be in its own context. That way, the test ensures that the exception is raised only in the exact statement where you expect it.

```

cooperlees (changes requested):
Thankyou both of you! Looks and sounds like a good bug to highlight to me. Lets just fix that suggestions of `regexp` that seems to not exist and I'll merge.

cooperlees (approved):

PR #385, opened 2023-05-22
Improve B033 (duplicate set items)

### LINKED ISSUE #384: B033 should say what the duplicate set item is
I got an error `taxonomy/parsing.py:87:11: B033 Sets should not contain duplicate items. Duplicate items will be replaced with a single item at runtime.`

Now, part of this set looks like this:

```
            "İ",
            "À",
            "Á",
            "Ć",
            "Á",
            "Å",
            "Ä",
```

I think it's the `Á`, but it would have been more helpful if flake8-bugbear had told me directly.

I'll work on a fix.

JelleZijlstra:
- Include the repr() of the duplicate item in the error
- Point the error directly at the duplicate item

Fixes #384

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1356,7 +1356,9 @@ class BugBearVisitor(ast.NodeVisitor):
             if not isinstance(elt, ast.Constant):
                 continue
             if elt.value in seen:
-                self.errors.append(B033(elt.lineno, elt.col_offset, vars=(repr(elt.value),)))
+                self.errors.append(
+                    B033(elt.lineno, elt.col_offset, vars=(repr(elt.value),))
+                )
             else:
                 seen.add(elt.value)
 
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
LGTM - So we get an error per duplicate now right? So if the "multi_line" set had 0, False as well, we'd get two errors?

(Just confirming I'm reading it right)

I ask, cause if so that's awesome. I prefer tools that show you all the errors so you can fix them all and don't get new errors after you fix on where ever possible ...

JelleZijlstra:
Yes, we'll give errors for all of them. Could add a test showing that explicitly if you like.

cooperlees:
> Yes, we'll give errors for all of them. Could add a test showing that explicitly if you like.

Awesome. I won't block on it, so feel free to merge, but if you have time to do so, I think it clearly states the full intention and guards against regression in the future.

JelleZijlstra:
Easy enough to do, I pushed the new test

Issue #378, opened 2023-03-28
B905 should not be raised for itertools.repeat


nonamethanks:
```python
import itertools
list(zip([1, 2, 3], itertools.repeat("a")))
```

This should not raise a B905. `repeat` is by design infinite, that's the entire reason why it exists.

NewGlad:
can be closed

cooperlees:
Thanks!

PR #388, opened 2023-05-29
Add exclusions for B905

NewGlad:
Fix for #378


cooperlees (approved):
LGTM + tests show bugbear doing what we expect.

Issue #383, opened 2023-05-10
sdist is missing tox.ini


mtelka:
The sdist package at PyPI is missing `tox.ini`.  Please add the missing `tox.ini` file to sdist to make downstream testing easier.  Thank you.

cooperlees:
Sure. PR welcome or I'll eventually get to it. Should just be a `pyproject.toml` addition from memory ...

PR #389, opened 2023-06-05
add `tox.ini` in `sdist`

kyoto7250:
close https://github.com/PyCQA/flake8-bugbear/issues/383

This PR adds `tox.ini` in `MANIFEST.in` for downstream testing.


cooperlees (approved):
Thanks - forgot to get around to this! Will try get a release out today too cause I think we have another PR to get out.

I wonder if this can be all moved to pyproject.toml yet ... but that's for another day and another PR.

PR #392, opened 2023-06-26
Remove use of the deprecated `ast.Constant.s` attribute

AlexWaygood:
The `.s` attribute on `ast.Constant` nodes is an alias to the `.value` attribute. The `.s` alias has been deprecated since Python 3.8, and causes a DeprecationWarning to be emitted on Python 3.12


JelleZijlstra:
Thanks!

PR #398, opened 2023-07-01
Add B034: re.sub/subn/split must pass flags/count/maxsplit as keyword arguments

### LINKED ISSUE #397: New check: re.sub[n]/re.split with regex flag as third argument w/o keyword
### Context
* https://github.com/python/cpython/issues/99918#issuecomment-1333382582
* https://github.com/python/cpython/issues/56166

### Example code
```python
$ re.sub('a', 'b', 'aaa')
'bbb'
$ re.sub('a', 'b', 'aaa', re.IGNORECASE)
'bba'
>>> re.split(' ', 'a a a a')
['a', 'a', 'a', 'a']
>>> re.split(' ', 'a a a a', re.I)
['a', 'a', 'a a']
>>> re.split(' ', 'a a a a', flags=re.I)
['a', 'a', 'a', 'a']
```

The fourth parameter to `re.sub[n]` is in fact `count`, and the third to `split` is `maxsplit`, and as seen in the above linked issues this is a *very* common mistake.

### Implementation
The check would consist of checking calls to `re.split`, `re.sub` and `re.subn`, with three or more parameters, where the third parameter does not have an explicit keyword.

To make it more lenient, it could require the third param to match re.[uppercase], or match any all-caps variable to cover `from re import IGNORECASE`, or any variable at all to cover `myregexflags = re.I | re.X`.
The WIP cpython PR's seem to be on track to enforcing the 3rd and 4th parameters to be kw-only though, and in case that actually becomes deprecated by 3.12 then the maximally strict variant of simply enforcing that when running <3.12 seems clearly appropriate.

jakkdl:
Fixes #397

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -188,7 +188,7 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B033**: Sets should not contain duplicate items. Duplicate items will be replaced with a single item at runtime.
 
-**B034**: Calls to `re.sub`, `re.subn` or `re.split` should pass `flags` or `output`/`maxsplit` as keyword arguments. Not doing so is a very common source of confusion, and will likely be deprecated in future python versions.
+**B034**: Calls to `re.sub`, `re.subn` or `re.split` should pass `flags` or `output`/`maxsplit` as keyword arguments. Not doing so is a very common source of confusion.
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -188,7 +188,7 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B033**: Sets should not contain duplicate items. Duplicate items will be replaced with a single item at runtime.
 
-**B034**: Calls to `re.sub`, `re.subn` or `re.split` should pass `flags` or `output`/`maxsplit` as keyword arguments. Not doing so is a very common source of confusion.
+**B034**: Calls to `re.sub`, `re.subn` or `re.split` should pass `flags` or `count`/`maxsplit` as keyword arguments. It is commonly assumed that `flags` is the third positional parameter, forgetting about `count`/`maxsplit`, since many other `re` module functions are of the form `f(pattern, string, flags)`.
 
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


tests/test_bugbear.py
```diff
@@ -520,6 +538,7 @@ def test_b908(self):
         )
         self.assertEqual(errors, expected)
 
+    @unittest.skipIf(sys.version_info < (3, 8), "not implemented for <3.8")
```
JelleZijlstra, line 541: We no longer support 3.7

jakkdl: oh shoot, messed up a rebase. thanks

JelleZijlstra (approved):

README.rst
```diff
@@ -188,6 +188,8 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B033**: Sets should not contain duplicate items. Duplicate items will be replaced with a single item at runtime.
 
+**B034**: Calls to `re.sub`, `re.subn` or `re.split` should pass `flags` or `output`/`maxsplit` as keyword arguments. Not doing so is a very common source of confusion.
```
cooperlees, line 191: I had to go read the issue to understand what the confusion was here. Let’s try make this clearer to save people hunting the issue down.
```suggestion
**B034**: Calls to `re.sub`, `re.subn` or `re.split` should pass `flags` or `output`/`maxsplit` as keyword arguments. Not doing so is a very common source of confusion due to the positions of these arguments differing in the `re` module APIs.
```
- Happy for better wording.

bugbear.py
```diff
@@ -1770,6 +1792,12 @@ def visit_Lambda(self, node):
     )
 )
 
+B034 = Error(
+    message=(
+        "B034 {} should pass `{}` and `flags` as keyword arguments to avoid confusion."
```
cooperlees, line 1797:
```suggestion
        "B034 {} should pass `{}` and `flags` as keyword arguments to avoid confusion due to different argument positions."
```

cooperlees (changes requested):
Let’s make readme and maybe the output message a little clearer on why we’re suggestion this change. I feel lots of bugbear users might be upset with this one if we don’t make the error messages as helpful as they could be.

README.rst
jakkdl: Isn't it just as much, if not more, about people not being aware that `count`/`maxsplit` are arguments at all, and just being used to passing `flags` as the third argument to all the other functions in `re`?
I agree with making it more clear though

cooperlees (approved):

PR #403, opened 2023-07-21
docs: fix grammar "belong to"

tanducmai:


JelleZijlstra:
"belong in" is correct here. See e.g. https://ell.stackexchange.com/questions/32375/what-is-the-difference-between-belong-in-and-belong-to

PR #406, opened 2023-09-02
fix name collision for node_stack on python 3.12

jakkdl:
Noticed this failing in #405. I don't really understand why it's defined both as a class member and a property, but removing it has no effect as far as I can tell. I think this only fails on 3.12 and as such might be unintended upstream.


jakkdl:
git blame shows these as quite old lines, 4980cc56 added the class member and 9e14a8c9 added the property

@Zac-HD

Zac-HD (approved):
Yep, looks like we just missed this (by one line in the diff!) when reviewing #191.  Thanks for the patch!

PR #410, opened 2023-09-13
Upgrade black

hauntsaninja:
Fixes #409


cooperlees (approved):
Thanks, but don't burn yourself out! I would have eventually done this. But appreciate it.

PR #415, opened 2023-09-21
B902: Add exceptions for standard library metaclasses

henzef:
I faced a false positive when deriving my metaclass from ABCMeta, which should be fixed with this commit. I also added EnumMeta to this list, because it was the only other public metaclass in the standard library.

### DIFF
--- a/tests/b024.py
+++ b/tests/b024.py
@@ -92,7 +92,7 @@ class multi_super_2(notabc.ABC, metaclass=abc.ABCMeta):  # safe
 
 
 class non_keyword_abcmeta_1(ABCMeta):  # safe
-    def method(self):
+    def method(cls):
         foo()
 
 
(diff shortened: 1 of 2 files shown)


cooperlees (changes requested):
I'm not sure of the exact cause here, but CI is unhappy ...

henzef:
I rebased the commit to the latest main branch in the hope that it would retrigger the CI runs, but that didn't work. Now I can no longer see the origin CI results :(

tests/b024.py
```diff
@@ -92,7 +92,7 @@ def method(self):
 
 
 class non_keyword_abcmeta_1(ABCMeta):  # safe
-    def method(self):
+    def method(cls):
```
cooperlees, line 95: Is there any value to have both method(self) and method2(cls)?

cooperlees (approved):
Thanks. A worthy enhancement.

PR #424, opened 2023-11-23
#418: Modify B017 to no longer have a false negative when raises() is imported directly from pytest

### LINKED ISSUE #418: B017: False negative when "from" imports used
## Summary
B017 fails when raises is imported directly

## Version
Replicated on version `23.9.16`

## Sample code
```
import pytest
from pytest import raises


def foo():
    raise


def test_pytest_raises():
    with raises(Exception):
        foo()
    with pytest.raises(Exception):
        foo()
```

azinneck0485:
Logic for check B017 has been updated to check for cases where `with raises()` is used when `raises` is imported using an `ast.ImportFrom` node (that is, `from pytest import raises`) rather than an `ast.Import` node, which the current logic handles.

This has included some updates for check B005 to ensure `ast.ImportFrom` nodes are added to the "imports" set maintained in the `BugBearVisitor` class.

Tests have been updated to check that this code pattern is now flagged with B017 and tox tests have passed for Python 3.11 (which is what I have installed).

Fixes #418.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -654,25 +654,28 @@ class BugBearVisitor(ast.NodeVisitor):
         lookup.
         """
         item = node.items[0]
-        item_context = item.context_expr
-
-        if (isinstance(item_context.func, ast.Name)
-            and item_context.func.id == "raises"#:  # ast.Name element with id = raises implies a call to a "raises" method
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Many thanks for this PR. Code looks good and tests cover all new functionality. Sorry about being slow.

Issue #391, opened 2023-06-09
Warn about static key in a dict comprehension


Skylion007:
The full case for using this is made in this ruff issue. I can't think of any case where using a static key in a dict comprehension is intended unless there is some recursive reduction with the old value.

cooperlees:
Added with attached PR

PR #426, opened 2023-11-26
implement B035 check for static keys in dict-comprehension

r-downing:
Implementing static-key check for dict comprehension from https://github.com/PyCQA/flake8-bugbear/issues/391

This not only checks for hard-coded consts, but also variables that likely aren't changing as part of the generator expression


bugbear.py
```diff
@@ -954,6 +955,36 @@ def check_for_b031(self, loop_node):  # noqa: C901
                                 B031(node.lineno, node.col_offset, vars=(node.id,))
                             )
 
+    def _get_names_from_tuple(self, node: ast.Tuple):
+        for dim in node.dims:
```
cooperlees, line 959: Is this >= 3.9 only?

r-downing: yup, fixed. Sorry, didn't have actions enabled on my fork, got that working now too

cooperlees (approved):
Awesome. Many thanks for this. Check totally makes sense.

Gonna sync us to latest black and get a release out tonight.

PR #427, opened 2023-11-27
add flake8 w/ bugbear to pre-commit config

r-downing:
For https://github.com/PyCQA/flake8-bugbear/issues/412

Just adds basic flake8 +bugbear check to pre-commit config

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,4 +19,4 @@ repos:
     hooks:
     - id: flake8
       additional_dependencies: [flake8-bugbear]
-      exclude: ^tests/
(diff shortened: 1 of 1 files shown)


r-downing:
Also, I noticed that the pre-commit auto-updating is breaking the test-suites?

We could add a similar `exclude: ^tests/b.*` line to pre-commit for black and isort and to bypass the formatting and check for these specific files, since they depend on line and column numbers and contain examples of bad code

cooperlees (approved):
This will at least stop us failing our own checks that are released :)

jakkdl:
:tada:

PR #428, opened 2023-11-27
update black version, and exclude the tests/b*** files from formatting checks so they don't break tests

r-downing:
Fix to prevent pre-commit ci autoupdate from breaking tests by reformatting test-code and changing line nums. Example: https://github.com/PyCQA/flake8-bugbear/pull/422/files

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,10 +6,12 @@ repos:
     rev: 5.12.0
     hooks:
       - id: isort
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Rightio - This makes life easier.

I was having black format the test code as we use this for signal with releasing black itself ... So more code the better, but it's very little signal ... So happy to do this - Thanks.

PR #430, opened 2023-11-28
B035: Fix false positive when named expressions are used (#429)

### LINKED ISSUE #429: B035: False positive for comprehensions that use a walrus operator
```python
        by_payment = {
            payment_id: charge
            for charge in charges
            if (payment_id := charge.metadata.get('payment_id')) is not None
        }
```
This results in a B035, even though `payment_id` is definitely not a static value, it's derived from the individual `charge` object.

I'm not surprised this wasn't caught, because comprehensions and if expressions are some of the few exceptions where the use of a variable can occur before its definition, so it requires special casing when looking up names.

So the error is in in this section of the check
```python
        elif isinstance(node.key, ast.Name):
            if node.key.id not in self._get_dict_comp_loop_var_names(node):
                self.errors.append(
                    B035(node.key.lineno, node.key.col_offset, vars=(node.key.id,))
                )
```
Instead of only looking for the loop variable names in the `target` node it also needs to look for any `ast.NamedExpr` within the `ifs` nodes of each `ast.comprehension`.

Daverball:
Fixes #429

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -962,12 +962,17 @@ class BugBearVisitor(ast.NodeVisitor):
             elif isinstance(dim, ast.Tuple):
                 yield from self._get_names_from_tuple(dim)
 
-    def _get_dict_comp_loop_var_names(self, node: ast.DictComp):
+    def _get_dict_comp_loop_and_named_expr_var_names(self, node: ast.DictComp):
+        finder = NamedExprFinder()
         for gen in node.generators:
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

cooperlees:
Very nice, thanks. This is indeed something that didn't cross my mind at all ... I'll cut a release later today to avoid others hitting this.

PR #432, opened 2023-11-30
include tuples in b018 useless-statement check

r-downing:
I was looking for something to detect useless tuple statements, noticed this had a useless-statement check but didn't include tuples. Any reason for that? Otherwise, figured we could add it?


tests/test_bugbear.py
```diff
@@ -270,6 +270,8 @@ def test_b018_functions(self):
         expected = [B018(line, 4) for line in range(15, 25)]
         expected.append(B018(28, 4))
         expected.append(B018(31, 4))
+        expected.append(B018(32, 4))
+        expected.append(B018(33, 4))
```
cooperlees, line 274: Why do we not need 34 here when we include 32 for modules tests? What am I missing here, cause the variable assign happens in a function it's fine?

cooperlees (changes requested):
Read again this morning and still confused at some of the test results ... I feel I'm missing something - Can you explain my missing context please :)

tests/test_bugbear.py
r-downing: Oh it's a little confusing because the line numbers for the three different test cases are referring to the 3 different `b018_....py` files (but the line numbers are _almost_ the same for 2 of them. Added the same 3 lines to each of the files and corresponding test cases, in all cases the first 2 lines are bad and the 3rd is good

```python
    (1,)  # bad
    (2, 3)  # bad
    t = (4, 5)  # good
```

r-downing:
Something else to note here - a dangling comma after a single thing (e.g. a function call) will turn it into a tuple, which would get flagged as a useless statement:

```python
print("abc")  # this is fine
```

```python
print("abc"),  # the comma essentially makes this a tuple (None,) so this would be flagged as a useless expression.
```

^ wondering if this might be confusing and should be labelled a little differently? Or even as separate error? E.g. `useless tuple expression - suggest removing the comma`

cooperlees (approved):

PR #433, opened 2023-12-01
add classname to b018 useless-expression output

r-downing:
Had another idea for an enhancement to the B018 warning, which might make things clearer, especially for the case of tuples. This adds the AST class name to the output. 

test file:
image

before and after:
image


cooperlees (approved):
O, very slick. Great idea! Thanks!

PR #434, opened 2023-12-02
B018 to detect useless-statements at all levels

r-downing:
The B018 useless-statement checker was only running at the top level of modules, classes, and functions. This meant that it wouldn't detect things inside any nested blocks - e.g. `if`, `for`, `while`, `try`, etc. Did a minor refactor to make the checker run during the basic visit function so it hits things in all scopes.

Example. See added test file for more

```python
[1, 2, 3]  # useless list - would catch this

if x:
    [1, 2, 3]  # useless list - would not catch this
```

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -370,6 +370,8 @@ class BugBearVisitor(ast.NodeVisitor):
         if is_contextful:
             self.contexts.pop()
 
+        self.check_for_b018(node)
+
     def visit_ExceptHandler(self, node):
         if node.type is None:
             self.errors.append(B001(node.lineno, node.col_offset))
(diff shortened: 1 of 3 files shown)


bugbear.py
```diff
@@ -1164,31 +1163,30 @@ def check_for_b903(self, node):
         self.errors.append(B903(node.lineno, node.col_offset))
 
     def check_for_b018(self, node):
```
r-downing, line 1165: this is mostly the same - just removing the toplevel for loop and de-indenting, renaming loopvar `subnode` to `node`. Can turn on "ignore-whitespace" option on PR view for easier review

Issue #156, opened 2021-02-04
Proposed Check: Unintentional str concat


jamesstidard:
One of the common mistakes I can make is to miss out a comma in collections of literal strings, resulting in the strings being concatenated.

```
my_list = ["a", "b", "c" "d"]  # note the missing comma between 'c' and 'd'
print(my_list)  # ["a", "b", "cd"]
```

I've not come across a time I've ever used this intentionally, and will either explicitly write `"c" + "d"` or use a format string `"{a} {b}` when `a` and `b` are variables.

If you intend this using brackets might be a fine way. `my_list = ["a", "b", ("c" "d")]`. Similar to the trailing comma tuple mistake where [flake8-comma](https://pypi.org/project/flake8-commas/) flags `my_int = 5,` but allows the more explicit `my_int = (5,)`.

The above would mean multiline strings could be concerted still as:

```
my_multiline = (
    "part one "
    "part two"
) 
```

As soon as a comma is introduced though it becomes a tuple, and I'd argue each str literal should have a comma between it or none should.

cooperlees:
I have also personally hit this and would love this added to bugbear.

I like the using parens "()" to make this non + concatting explicit, although I would prefer people just always use '+'. I bet this will cause a lot of noise to our users but will find bugs.

If you have the cycles, I'd take a PR, but would request very good unit testing here to ensure we're covering all of our edge cases please!

kkirsche:
+1 for me

In the past, I've used https://github.com/Instagram/Fixit to identify things like this. Reducing the tools necessary would be great, and they may have an example implementation that can be used as a foundation.

Legacy rule they had for preventing implicit concatenation:
https://github.com/Instagram/Fixit/blob/c3f8ff9e6cfa2199c6154eefea24ea39e5278dc6/legacy/rules/no_implicit_concat.py

PR #435, opened 2023-12-08
Detect inline implicit string concatenation

r-downing:
As specified here https://github.com/PyCQA/flake8-bugbear/issues/156 - this adds a check for inline implicit string concatenation as a new error B036

E.g.

```python
["a", "b" "c"]
```

becomes

```python
["a", "bc"]
```

This grabs the segment of the original source and parses it with libcst to see if it's a `ConcatenatedString`. Had to modify some unittests that manually fed in the AST to also feed in the source lines.


bugbear.py
```diff
@@ -372,6 +374,31 @@ def visit(self, node):
 
         self.check_for_b018(node)
 
+    def get_source_segment(self, node: ast.AST):
+        lines = self.lines
+        return lines[node.lineno - 1][node.col_offset : node.end_col_offset]
```
r-downing, line 379:
```suggestion
        return self.lines[node.lineno - 1][node.col_offset : node.end_col_offset]
```

cooperlees:
My only worry here is libcst is a big depedency to add ... We been light on (and could prob remove attrs with dataclasses these days - but I forget what we use) dependencies on purpose and wanted to be a pure python project for runtime simplicity. libcst is compiled but the right library for this job none the less.

I'd like to see how other contributors + maintainers feel about bringing this dependency in. It brings in many benefits, but now could slow our move to new Python runtimes and brings in the potential for people having wheel install issues + build the extension etc. etc.

JelleZijlstra:
I agree with your intuition that it's better to keep flake8-bugbear pure Python. It's good to minimize dependencies.

We could reconsider if there are more lints that we want to add but that are only achievable with libcst.

cooperlees:
Is there a feasible pure python way to implement these checks? If not, I think we should close this PR. I thank you for the efforts tho. Jelle and I (plus other maintainers) bear the pain of black and other projects having compiled dependencies. They lead to a lot of Issues opened. As it's really just me and Jelle here, I'd like to avoid that overhead.

I wonder if someone has started a dedicated libcst based flake8-plugin so people can choose a compiled plugin if they want. If so, maybe the check can go there.

Issue #174, opened 2021-06-29
Suggestion: New lint to forbid `except BaseException:`


untitaker:
This goes further than the existing lint against `except:`. The rationale is that in many codebases (celery task bodies and request handlers, in our case) it's extremely unlikely that `except BaseException:` is ever a valid way to handle errors. In our org we've seen people (usually ones who are not writing Python daily) introduce `except BaseException:` as per a suggestion from the lint against `except:` without diving deeper into the topic, and causing all kinds of problems. With such a new lint forbidding `BaseException` as well they'd be forced to think more about what exceptions they really want to handle, and hopefully they'll reconsider when writing `except (Exception, SystemExit, KeyboardInterrupt, ...)`.

If y'all would accept such an opt-in lint I'll start working on it immediately.

cooperlees:
I agree this is dangerous and never needed. I am even open to having the release this makes it into default to warning on this. What do others think here. Please speak up if you don't want this on by default.

Please just have good unittests + nice explanation in the errors raised + README.md. Bonus points, add to CHANGES to please.

Zac-HD:
It's not *easy* to check that every code path in the except-block re-raises the exception, but [we've done it in flake8-trio](https://github.com/Zac-HD/flake8-trio/blob/5072c7d9e4afa340b94ef98fa901afdd563cac65/flake8_trio.py#L606-L608) fwiw.

r-downing:
I'm happy to implement this - would the consensus be default or opt-in? 

I was also looking for similar checks for plain `Exception` similar to pylint's `broad-exception-raised` and `broad-exception-caught` - these should probably be opt-in. Would be happy to implement these as well, any interest?

Zac-HD:
So long as we only trigger on blocks that don't re-raise, I think default enabled is reasonable.  Thanks for volunteering to help with this!

dejvidq:
Shouldn't this issue be closed after it was implemented in #438?

cooperlees:
Yes.

PR #438, opened 2023-12-13
Implement B036 check for `except BaseException` without re-raising

r-downing:
Started implementing this new check per https://github.com/PyCQA/flake8-bugbear/issues/174

Currently just looking for `BaseException` - I figured `Exception` should maybe be a separate, optional one.

It only checks for re-raises at the top-level of the except. I thought maybe checking all branches of of a complex `except` block to see they all re-raise might be overkill - WDYT?

This does contradict the advice (and failing test-cases) of B001:
> Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.Flake8(B001)

So those would also need to be updated, and should probably change that wording - what do you think should be suggested instead? Or just "use a more specific exception" or something like that?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -195,6 +195,8 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B035**: Found dict comprehension with a static key - either a constant value or variable not from the comprehension expression. This will result in a dict with a single key that was repeatedly overwritten.
 
+**B036**: Found ``except BaseException:`` without re-raising (no ``raise`` in the top-level of the ``except`` block). This catches all kinds of things (Exception, SystemExit, KeyboardInterrupt...) and may prevent a program from exiting as expected.
+
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -408,7 +408,8 @@ class BugBearVisitor(ast.NodeVisitor):
             if maybe_error is not None:
                 self.errors.append(maybe_error)
         if "BaseException" in names and not any(
-            isinstance(subnode, ast.Raise) and subnode.exc is None
+            isinstance(subnode, ast.Raise)
+            and (subnode.exc is None or subnode.exc.id == node.name)
             for subnode in node.body
         ):
             self.errors.append(B036(node.lineno, node.col_offset))
@@ -1677,8 +1678,8 @@ B001 = Error(
     message=(
         "B001 Do not use bare `except:`, it also catches unexpected "
         "events like memory errors, interrupts, system exit, and so on.  "
-        "Prefer `except Exception:`.  If you're sure what you're doing, "
-        "be explicit and write `except BaseException:`."
+        "Prefer excepting specific exceptions  If you're sure what you're "
+        "doing, be explicit and write `except BaseException:`."
     )
 )
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -324,6 +324,31 @@ def _typesafe_issubclass(cls, class_or_tuple):
         return False
 
 
+class ExceptBaseExceptionVisitor(ast.NodeVisitor):
+    def __init__(self, except_node: ast.ExceptHandler) -> None:
+        super().__init__()
+        self.root = except_node
+        self._re_raised = False
+
+    def re_raised(self) -> bool:
+        self.visit(self.root)
+        return self._re_raised
+
+    def visit_Raise(self, node: ast.Raise):
+        """If we find a corresponding `raise` or `raise e` where e was from
+        `except BaseException as e:` then we mark re_raised as True and can
+        stop scanning."""
+        if node.exc is None or node.exc.id == self.root.name:
+            self._re_raised = True
+            return
+        return super().generic_visit(node)
+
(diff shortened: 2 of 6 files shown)


README.rst
```diff
@@ -195,6 +195,8 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B035**: Found dict comprehension with a static key - either a constant value or variable not from the comprehension expression. This will result in a dict with a single key that was repeatedly overwritten.
 
+**B036**: Found ``except BaseException:`` without re-raising (no ``raise`` in the top-level of the ``except`` block). This catches all kinds of things (Exception, SystemExit, KeyboardInterrupt...) and may prevent a program from exiting as expected.
```
cooperlees, line 198: Do we also catch a bare `except` here and error?

bugbear.py
```diff
@@ -677,7 +683,7 @@ def check_for_b017(self, node):
                     and item_context.func.id == "raises"
                     and isinstance(item_context.func.ctx, ast.Load)
                     and "pytest.raises" in self._b005_imports
-                    and "match" not in [kwd.arg for kwd in item_context.keywords]
+                    and "match" not in (kwd.arg for kwd in item_context.keywords)
```
cooperlees, line 686: Just making tuples since they don't need to be immutable?

tests/b036.py
```diff
@@ -0,0 +1,35 @@
+
+try:
+    pass
+except BaseException:  # bad
+    print("aaa")
+    pass
```
cooperlees, line 6: Can we add and make this work easily? If not, no worries.

```suggestion
try:
    pass
except:  # bad
    print("aaa")
    pass
```

cooperlees (changes requested):
>  I figured Exception should maybe be a separate, optional one.

I agree here, this will be much more controversial.

> It only checks for re-raises at the top-level of the except. I thought maybe checking all branches of of a complex except block to see they all re-raise might be overkill - WDYT?

I'm happy to start simple here and we can add to it if people complain about to much noise ... But covering as many re-raises as we can will be key.

> So those would also need to be updated, and should probably change that wording - what do you think should be suggested instead? Or just "use a more specific exception" or something like that?

I'm happy with "please use a more specific exception" in the wording. Using BaseException shouldn't really be needed in any non core cpython/runtime libraries imo.

Please fix the unit tests. Did black break it?

tests/b036.py
r-downing: Look like bare except already caught by both flake8 and bugbear:

```python
try:
    pass
except:
    pass
```
gives both:
- B001 Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.
- E722 do not use bare 'except'

Should this be consolidated with B001, rather than a new error?

cooperlees: Hmm - that's a good question. I'd like that, but I wonder if that would upset people?

I vote yes, but lets see if any other contributors / maintainers weigh in here ...

bugbear.py
r-downing: Actually not a tuple, just parenthesizing the generator:

```
>>> x = [i for i in range(3)] 
>>> x
[0, 1, 2]
>>> x = (i for i in range(3))
>>> x
<generator object <genexpr> at 0x0000021C2DFCAA80>
```

Just a tiny efficiency improvement, rather than having iterate the generator to construct a list, then iterate through the list to find the thing... just iterate the generator directly

tests/b036.py
r-downing: I guess keeping separate also gives you the flexibility to enable/disable either one? Maybe that's the safer option?

r-downing:
> Please fix the unit tests. Did black break it?

Oh I think the unittests broken currently because the B001 test is now throwing new B036 errors in addition to the expected ones. I might change the "good" examples in the B001 test to something other than `BaseException` since that's the other thing we're trying to discourage

r-downing:
Ok fixed up the unittests and wording. Regarding finding the re-`raise` - I'm thinking I can just do a recursive search and see if there's a raise in **any** branch? Rather than the top-level? Shouldn't be too hard to implement...

I could imagine some weird intentional case where you might have something like

```python
except BaseException as e:
    if ...:
        raise e
    else:
        # don't raise
```

WDYT?

tests/b036.py
cooperlees: Yeah ok, I like this now the more I think about it.

cooperlees:
That sounds good to me for a first effort finding re-raises.

Did you check out what flake8-trio did that was mention on the issue? Could maybe steal some logic.

r-downing:
> Did you check out what flake8-trio did that was mention on the issue? Could maybe steal some logic.

Took a look at it, seems like they're trying to check **all** paths. I just added a basic recursive search for a corresponding `raise` on **any** path, to start with. I think this might make more sense actually, or at least be a little more forgiving.

It ignores `raise` within a nested except, as I think it should. Check out the b036.py for examples

cooperlees (approved):
Sounds a good start to me. Thanks for working through and iterating!

PR #439, opened 2023-12-15
make B017 also apply to BaseException

r-downing:
on the subject of `BaseException` I thought it might make sense for this `assertRaises` check to apply to that as well


cooperlees (approved):
Agree. This is worse, IMO.

Wonder how noisy this will be ... but i think it's a big enough one for people to want to clean up in their code bases ...

PR #442, opened 2023-12-19
Implement B037 check for yielding or returning values in __init__()

r-downing:
Implement check for ``return <value>``, ``yield``, ``yield <value>``, or ``yield from <value>`` in class ``__init__()`` method. No values should be returned or yielded, only bare ``return``s are ok.

These would be runtime errors, if and when these statements are reached.

This does trigger on `return None` which is equivalent to `return`, but kind of confusing and pointless to write in a function that shouldn't return anything.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -351,35 +351,6 @@ class ExceptBaseExceptionVisitor(ast.NodeVisitor):
         return super().generic_visit(node)
 
 
-class ClassInitVisitor(ast.NodeVisitor):
-    """Use get_errors() to scan a ClassDef's __init__() method to ensure that
-    it doesn't have any `yield` or `return <value>`"""
-
-    def __init__(self) -> None:
-        super().__init__()
-        self.errors: list[error] = []
-
-    @staticmethod
-    def get_errors(class_node: ast.ClassDef) -> list[error] | None:
-        funcs = (n for n in class_node.body if isinstance(n, ast.FunctionDef))
-        init_func = next((n for n in funcs if n.name == "__init__"), None)
-        if init_func is None:
-            return None
-        visitor = ClassInitVisitor()
-        visitor.visit(init_func)
-        return visitor.errors
-
-    def visit_Return(self, node: ast.Return) -> None:  # noqa: B906
-        if node.value is not None:
-            self.errors.append(B037(node.lineno, node.col_offset))
-
-    def visit_Yield(self, node: ast.Yield) -> None:  # noqa: B906
-        self.errors.append(B037(node.lineno, node.col_offset))
-
-    def visit_YieldFrom(self, node: ast.YieldFrom) -> None:  # noqa: B906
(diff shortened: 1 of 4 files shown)


tests/b037.py
```diff
@@ -0,0 +1,33 @@
...
+        pass
+
+    def __init__(self, k="") -> None:
+        yield from []  # bad
+
+
+class D(C):
+    def __init__(self, k="") -> None:
+        super().__init__(k)
+        return None  # bad
```
cooperlees, line 29: I can see people disliking this due to

```
Explicit is better than implicit
```
In the Zen of Python. 

Can we just check if a `None` is being explicitly returned and not error in that case since we allow `return` ...

cooperlees (changes requested):
LGTM - But I think I would like if we allowed `return None` unless it's difficult that I'm not seeing ...

I'm a big explicit is better than implicit person ...

tests/b037.py
r-downing: Yeah it's easy enough to do if you prefer, but my argument against allowing `return None` is that, while `return` is functionally equivalent to `return None` in python, the latter should only be used in cases where you _might_ return something (like a C function that returns either a valid pointer or `NULL`). Whereas just bare `return` is ok, when used as a early exit point from a function (like a C void function).

some quick examples:

```python
def get_thing() -> Thing | None:
    """return a thing if possible"""
    for thing in things:
        if condition:
            return thing
    return None  # this line could be omitted, but functions that have a non-empty return
                 # in some case should have non-empty return in all cases (pylint would 
                 # complain about this for example)
```

vs 

```python
class Example:
    def __init__(self, size, do_extra_stuff = False):
        self.x = ...

        if not do_extra_stuff:
            return  # early exit. Could do `return None` but that implies
                    # that we might return something somewhere else?

        # do some more stuff
```

let me know what you think, thanks!

bugbear.py
```diff
@@ -349,6 +351,35 @@ def visit_ExceptHandler(self, node: ast.ExceptHandler):
         return super().generic_visit(node)
 
 
+class ClassInitVisitor(ast.NodeVisitor):
```
JelleZijlstra, line 354: Might be cleaner to put this inside the `BugBearVisitor`

tests/b037.py
JelleZijlstra: I agree that it's better to lint against `return None`.

bugbear.py
r-downing: Done, LMK if that looks better

r-downing:
Waiting on anything else here?

JelleZijlstra (approved):

tests/b037.py
cooperlees: Since this is `__init__` only, I can live with it. I'll have code to 'fix' cause of it.

cooperlees (approved):

Issue #445, opened 2024-01-10
Rule to detect changes to iterable object of loop


mimre25:
Recently I've ran into a classic bug in a code base, to illustrate a minimal example:

```py
some_list = [1,2,3]
for i in some_list:
  if i % 2 == 0:
    some_list.remove(i)
  print(i)
```
To nobodies surprise this prints only 1 and 2.

I was surprised that none of my linters caught this.

Anyway, would it be possible to add a rule that detects changes to the iterable that is being iterated?

I'm also more than happy to contribute :slightly_smiling_face:

cooperlees:
Yeah, I think we've all been here.

This sounds good, but I think in general we should just detect if it is a mutable iterable when we see remove in a loop and suggest saving the indexs and then removing them in reverse? Or am I missing something here?

Anyways, in general this sounds good. Feel free to make a POC diff and we can comment on there.

mimre25:
I've went with an approach that alerts for any mutations of the loop iterator (eg `.remove` or `del`).

PR is in #446

mimre25:
FYI: While starting to implement this rule into `ruff` yesterday I might have found a bug in my initial implementation.

@cooperlees please hold off the release until I verify my suspicion and fix it ~ I'll check this out later today.

mimre25:
Ok, I was seeing something, but that code wasn't used - I've opened #488 to remove it, to avoid future bugs :slightly_smiling_face:

cooperlees:
This is released and we have dedicated improvements issue. So let's move there.

I yanked 24.1.15 + 24.1.15 due to bugs.

PR #446, opened 2024-01-12
feat(rules): Add rule to check for mutations of loop iterator

mimre25:
This implements a POC for #445.

I opted to go the route of alerting for any mutations of a loop iterator.

Note: 
I picked B999 temporarily, not sure how the numbers are picked.
Also, I'm very much open to change the error message.

Would be happy if you could give this a review @cooperlees  :slightly_smiling_face:


bugbear.py
```diff
@@ -495,6 +496,7 @@ def visit_For(self, node):
         self.check_for_b020(node)
         self.check_for_b023(node)
         self.check_for_b031(node)
+        self.check_for_b999(node)
```
cooperlees, line 499:
```suggestion
        self.check_for_b038(node)
```
We generally just go next number unless we make optional - They are >900.

bugbear.py
```diff
@@ -1544,6 +1546,18 @@ def check(num_args, param_name):
         elif node.func.attr == "split":
             check(2, "maxsplit")
 
+    def check_for_b999(self, node: ast.For):
```
cooperlees, line 1549:
```suggestion
    def check_for_b038(self, node: ast.For):
```

bugbear.py
```diff
@@ -2045,6 +2102,8 @@ def visit_Lambda(self, node):
         " statement."
     )
 )
+
 B950 = Error(message="B950 line too long ({} > {} characters)")
 
+B999 = Error(message="B999 editing loop iterable can cause unintended bugs!")
```
cooperlees, line 2108:
```suggestion
B999 = Error(message="B038 editing a loop's mutable iterable often leads to unexpected results/bugs")
```
I'm not tied to this, but just tried to be as explicit and descriptive as possible

cooperlees (changes requested):
Many thanks for this. Looks good. I like this starting point.

Can you please:
- Lets go with b038
- Add more detailed few lines explanation into READNE.rst

bugbear.py
mimre25: Sounds good!

cooperlees (approved):
Many thanks for your contribution!

Will get a release our next week.

sahuagin:
I have an example for a false positive (I hope), if it's desired.
Setup. A CSV file as input with the first item in each row being an index number.
```python
        id_pos = 0
        # lf is LogFile not LazyFrame
        lf = self._logger
        with self.rlock:
            lf.seek(0, 0)
            cpos = lf.tell()
            for line in lf:
                if len(line.strip()) == 0:
                    continue
                sl = line.split(b",", 1)
                self._mapper[sl[id_pos]] = cpos
                cpos = lf.tell()
```
Desired behavior is to find all indexes and record the file position where they start.
This just started getting flagged, which led me here. There are 4 more flags in the same file,
tell()
seek() x2
write()

JelleZijlstra:
@sahuagin thanks for the report, #453 should fix this.

PR #448, opened 2024-01-15
refactor(b038): Cleanup unused code in B038Checker

mimre25:
The unused code would introduce a bug when used, thus I believe it's better to remove it.

Essentially, it would've flagged _any_ `del` statement within a loop, without check for the name. 
However, because the `B038Checker` visits `Del` nodes, and doesn't call `generic_visit` from there, we'll never reach a `Name` node that has the `ctx == Del`.

I've added a test case, just to make sure :slightly_smiling_face:

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1623,11 +1623,6 @@ class B038Checker(ast.NodeVisitor):
 
         self.generic_visit(node)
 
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Awesome. Thanks for finding before it went wide and we got flooded with false positives!

PR #453, opened 2024-01-16
fix(b038): Restrict rule to mutating functions only

### LINKED ISSUE #451: B038 false positive in 24.1.15 + 24.1.16
First of all thanks for this new B038 new rule, it's a great addition!

Using flake8-bugbear `24.1.15`

The following code is flagged.
```python
some_list = [1, 2, 3]
for elem in some_list:
    print(elem)
    if elem == 2:
        found_idx = some_list.index(elem)  # should not error
        break
```

Although a method of `some_list` is called I believe this doesn't have any effect on the mutable

mimre25:
Previous implementation produced false positives.
This fixes those and adds some more tests.

Fixes #451 

@cooperlees this fixes the bugs described in the issue - There are some more mutating functions (like a ~`dict.popitem()`~ - ok that was an easy one, but `=` and `*=` and `+=` are still left to do) that I won't have time to add today. 
That said, this will only cause false negatives, so I figured I'd push this and create the PR so you can release a fix for the FPs.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1608,6 +1608,7 @@ class B038Checker(ast.NodeVisitor):
         "extend",
         "insert",
         "pop",
+        "popitem",
     )
 
     def __init__(self, name: str):
--- a/bugbear.py
+++ b/bugbear.py
@@ -1598,6 +1598,18 @@ def compose_call_path(node):
 
 
 class B038Checker(ast.NodeVisitor):
+    # https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types
+    MUTATING_FUNCTIONS = (
+        "append",
+        "sort",
+        "reverse",
+        "remove",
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Sounds good to me and sage way to start hardening it.

Will let a few more eyes see it, and release later today.

Appreciate the prompt action, considering I didn't let you know when I was releasing it. I should do that moving forward too with new contributions like this.

mimre25:
No worries, I just had an hour in between things just now and this perfectly filled it :slightly_smiling_face: 

I will look into a good way to harden it further tomorrow and write up a proposal in an issue then.

JelleZijlstra (approved):

mimre25:
FYI, I've outlined the proposal on how to harden the rule of B038 further in #454

PR #456, opened 2024-01-22
Change B038 to B909 and make it optional

### LINKED ISSUE #455: B038 false positives
having the following code;

```
 mylist = list(range(5))
 for item in mylist:
    if item == 2:
        mylist.remove(item)  # <- badly detects B038 here
        break
```
bugbear badly detects B038.

Sample: https://doc.wikimedia.org/pywikibot/master/_modules/proofreadpage.html#ProofreadPage
in index property (search for `def index`). As B038 is descibed as _Found a mutation of a mutable loop iterable inside the loop body. Changes to the iterable of a loop such as calls to list.remove() or via del **can cause** unintended bugs._ This rule should be an optionated warning instead of a general rule.

mimre25:
While writing this up, I was thinking about making this optional and have two questions:
1. Should we change the B code (as I did)?
2. What would you say are the guidelines for making a rule optional?

---

B038 lead to some false positives that stem from methods defined in the standard library that have the same name as mutating functions for container types like lists and dicts.
Thus we decided to make this rule optional.
See https://github.com/PyCQA/flake8-bugbear/issues/455 for the related discussion.

Closes #455


cooperlees:
> Should we change the B code (as I did)?
Yes, but lets mark 38 as to skip / not use as people might have it already blocked in their configs now ...

> What would you say are the guidelines for making a rule optional?
Good question, we've always just been adhoc. Generally if there's a good chance majority of people don't need / want / the lint will be too noisey unless helped / tweaked it should be optional.

Happy for others to chime in with a more concrete definition. We can add it into the README etc. if we want too.

I'll try review this today..

cooperlees (approved):
LGTM - I just wonder if we should care about 'denylisting' b038 or just try to remember to not use it for next new check?

FozzieHi:
> LGTM - I just wonder if we should care about 'denylisting' b038 or just try to remember to not use it for next new check?

When B028 was renamed to B907, we reused B028. Not sure if we can get any analytics on whether that was the correct decision or not (i.e. how many people still have B028 in their ignore list), but I definitely think that if we decide to skip it, it should be included in the code base without any implementation so we don't forget.

FozzieHi:
Just noticed this when looking at this PR, but B907 isn't included in the `disabled_by_default` list for some reason. Maybe it was forgotten during the rename, should we just throw that in for this PR?

I'm curious at how the noise was reduced when it isn't even in the list, is it somehow still disabled anyway?

JelleZijlstra:
I'd prefer to not reuse the code, we're in no danger of running out of error codes. We can introduce some mechanism in the code to make sure the code stays reserved.

FozzieHi:
> Just noticed this when looking at this PR, but B907 isn't included in the `disabled_by_default` list for some reason. Maybe it was forgotten during the rename, should we just throw that in for this PR?
> 
> I'm curious at how the noise was reduced when it isn't even in the list, is it somehow still disabled anyway?

I've looked into this some more, and it seems like the [`should_warn` function](https://github.com/PyCQA/flake8-bugbear/blob/608df8407348843f2eb449e9c0852bf26141e378/bugbear.py#L171) is preventing all B9 rules from warning by default. The function itself is a workaround (see the function comments), so I still think it's a good idea to add all of the B9 rules manually to the `disabled_by_default` list.

Issue #413, opened 2023-09-18
B902: False positive when using the attrs library


henzef:
The adaptions to B902 cause a regression when using the attrs library:

    import attrs
    from typing import Any
    
    
    @attrs.frozen
    class SpamAndEggs:
        foobar: list[int] = attrs.field()
    
        @foobar.validator
        def _positive(
            self,
            attribute: "attrs.Attribute[list[int]]",
            value: Any,
        ) -> None:
            if any(i < 0 for i in value):
                raise ValueError()
    
    
    SpamAndEggs([1, 2, 3])

The error is:

    myfile.py:11:9: B902 Invalid first argument 'self' used for class method. Use the canonical first argument name in methods, i.e. cls.

My suggestion would be to remove validator and root_validator from the default list and allow pydantic user to set them via the commandline flag. I would imagine that other libraries also have @validator decorators or that some people write their own

henzef:
See #405

henzef:
Maybe @jakkdl or @Zac-HD have an idea how to fix it? I think attrs is a pretty common library, so it would be great if someone could fix it.

jakkdl:
With `--classmethod-decorators` you can overwrite the default values, so if you use `--classmethod-decoraters=classmethod` you'll suppress the errors.

The question is mostly what the default should be, but could maybe also update the documentation.

mcmasn-amzn:
I ran into this as well. I think a sensible default is just `classmethod`. If the checker can also include the python package of the decorator, then it makes sense to include some additional defaults.  For instance, `@validator` in attrs is not a classmethod, but it is in Pydantic.

jakkdl:
Being conservative with the defaults is probably the way to go, feel free to open a PR changing it + the docs. It should be quite painless to implement.

natemcmaster:
https://github.com/PyCQA/flake8-bugbear/pull/459

PR #459, opened 2024-02-05
Remove decorators named validator and root_validator from B902 checks

natemcmaster:
## What
This removes methods named `validator` and `root_validator` from the B902 check

## Why
Fixes https://github.com/PyCQA/flake8-bugbear/issues/413

Removes false positives for common cases, like using attrs validators

## Testing
ran `tox` and all tests are passing

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -39,7 +39,7 @@ python3 -m venv /path/to/venv
 
 ```console
 cd flake8-bugbear
-/path/to/venv/bin/pip install -e .[dev]
+/path/to/venv/bin/pip install -e '.[dev]'
 ```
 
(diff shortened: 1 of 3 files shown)


jakkdl (approved):
In hindsight I should've added methods to `tests/b902.py` with `@validator` and `@root_validator` to check that the defaults were properly parsed. You could perhaps add those now and mark them as failing in `tests.test_bugbear.test_b902`, but /shrug.

If I could approve the test run I would. I know @Zac-HD likes championing for only requiring approval for test runs if first-time contributors are new to github, as opposed to default where all first-time project contributors require approvals.

cooperlees (approved):
Let's make it so.

Issue #454, opened 2024-01-17
B038 improvements


mimre25:
# Description
With the introduction of B038 (#445), I initially introduced a bug causing many false positives, which should be fixed with #453.
#453 implements a check for the methods listed in [Python Builit-in Mutable Sequence Types documentation](https://docs.python.org/3/library/stdtypes.html#typesseq-mutable), but not the operators.
However, #453 left out some mutations for the sake of getting a fix to the flood of false positives (#451) out asap.

In this issue, I will propose further changes to harder B038, to detect more (hopefully all) mutations to iterables that the python builtins allow.

### Assignment Operations
This is a very obvious one that was not implemented yet, and I wonder how I could've overseen it.
This part will cover the following operations:
```py
for _ in foo:
  foo = bar
  foo *= bar
  foo += bar
  foo[i] = bar
  foo[i:j] = bar
  foo[i:j:k] = bar
  foo |= bar
  foo &= bar
  foo -= bar
  foo ^= bar
```
The implementation will check if any [`ast.Assign`](https://docs.python.org/3/library/ast.html#ast.Assign) and [`ast.AugAssign`](https://docs.python.org/3/library/ast.html#ast.AugAssign) have the name or a subscript of the name of the iterable as target.

Further, to also check for annotated assignments, the plan is to inspect [`ast.AnnAssign`](https://docs.python.org/3/library/ast.html#ast.AnnAssign) nodes that are not simple (ie have `simple = 0`). 
`simple = 1` are nodes of the form `foo: bar`, whereas `simple = 0` have any of the following forms:

```py
for _ in foo:
  foo: t = bar
  foo: t *= bar
  foo: t += bar
  foo[i]: t = bar
  foo[i:j]: t = bar
  foo[i:j:k]: t = bar
  foo: t |= bar
  foo: t &= bar
  foo: t -= bar
  foo: t ^= bar
```
I'm not sure if all of the above are legal syntax, but the implementation will likely look very similar and thus would cover all those cases out of the box.

Further, everything of the above also applies to object attributes, ie, we'd also cover `foo.bar = `-esque assignments if `foo.bar` is the loop iterable.

### Container-Specific Operations
#451 already implements a check for 2 container specific operations: [`list.sort()`](https://docs.python.org/3/library/stdtypes.html#list.sort) and [`dict.popitem()` ](https://docs.python.org/3/library/stdtypes.html#dict.popitem).
Further operations are missing:
[...]

PR #460, opened 2024-02-08
B909 improvements

mimre25:
image

ahh... nothing better than the classic "next week" that turns into a month :sweat_smile: 

This implements all improvements listed in #454.

I've implemented a rather "simple" check for unconditional breaks. 
I believe if we want to catch all scenarios, we'll eventually end up with the reachability problem, at least it feels like we're getting close to it.


cooperlees (approved):
Thanks for getting around to it! All looks good to me using tests to see what we flag on.

Issue #467, opened 2024-04-22
B909 has several false positives on black codebase


hauntsaninja:
E.g.
```
def lib2to3_parse():
    grammars = get_grammars()
    errors = {}
    for grammar in grammars:
        errors[grammar.version] = InvalidInput()
```

See https://github.com/psf/black/actions/runs/8776854301/job/24081105618?pr=4318 for more

cc @mimre25 in case interested in improving the check :-)

KumoLiu:
Same issue here:
https://github.com/Project-MONAI/MONAI/actions/runs/8777993558/job/24083761562#step:2:2647

https://github.com/Project-MONAI/MONAI/blob/03a5fa695ad02fcb916d5495e84f8f01883806e2/monai/apps/auto3dseg/auto_runner.py#L774

mimre25:
Thanks for tagging me! 

I'll take a look later this week and see if I can fix it right away, or if it blows up the scope (I doubt the latter one). :slightly_smiling_face:

gothicVI:
Hi,

we're seeing complaints for code like
```python
lst: list[dict] = [dic1, ..., dicn]
for dic in lst:
    dic["key"] = False
```

Is this related?

mimre25:
Potentially. I'm actually working on it right now, so I'll also check this.

mimre25:
I've created #469  - the fix was rather easy and simple, just an oversight in the initial implementation :grimacing:

I've ran it on the black code base with the only hit being [`./black/src/blib2to3/pgen2/pgen.py:235:21`](https://github.com/psf/black/blob/ba88fc372eaed34abb33ea02d6860b185388d928/src/blib2to3/pgen2/pgen.py?plain=1#L223-L236), which is a true positive considering the comment in line 223 (`for state in states:  # NB states grows while we're iterating`).

With Project-MONAI, I could also eliminate the cases pointed out, however, there is a discussion point (hence the PR only being a draft):

Should B909 allow altering an element via it's dictionary key? For example:
```py
some_dict = {...}
for key in some_dict:
    some_dict[key] = 3
    del some_dict[key]
```
Should this fire B909 or not? What do you think?

cc @cooperlees

gothicVI:
> Should B909 allow altering an element via it's dictionary key? For example:
> 
> ```python
> some_dict = {...}
> for key in some_dict:
>     some_dict[key] = 3
>     del some_dict[key]
> ```

[...]

PR #469, opened 2024-04-25
fix(b909): Fix false positive affecting containers of mutables

mimre25:
The false positives occurred when trying to edit a dictionary while iterating over a list of dictionaries:

```
lst: list[dict] = [{}, {}, {}]
for dic in lst:
    dic["key"] = False # was false positive - fixed now
```

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1631,7 +1631,10 @@ class B909Checker(ast.NodeVisitor):
 
     def visit_Assign(self, node: ast.Assign):
         for target in node.targets:
-            if isinstance(target, ast.Subscript) and _to_name_str(target.value):
+            if (
+                isinstance(target, ast.Subscript)
+                and _to_name_str(target.value) == self.name
+            ):
                 self.mutations[self._conditional_block].append(node)
         self.generic_visit(node)
 
--- a/tests/b909.py
+++ b/tests/b909.py
@@ -127,3 +127,7 @@ for _ in foo:
         bar.remove(1)
         break
     break
+
+lst: list[dict] = [{}, {}, {}]
+for dic in lst:
+    dic["key"] = False


henzef:
With this patch flake8 complains about the following construct in my code:

    for key in my_dictionary:
        my_dictionary[key] = True

which is safe to do, because it doesn't add or remove any keys. At least as far as I understand, please correct me if I am wrong.

cooperlees:
Thanks for this. What's the plan, did you want to add this fix or do you have more plans for this PR?

mimre25:
I'd be happy to add the fix for 
```py
for key in my_dictionary:
    my_dictionary[key] = True
```

at the time I opened the draft PR I wasn't 100% sure this is desired, but judging from the discussions in #467 and the :+1: on henzef [comment](https://github.com/PyCQA/flake8-bugbear/pull/469#issuecomment-2077316024) it seems like this is something we want. 

I'll hopefully get around to adding this later today.

mimre25:
This is ready now :slightly_smiling_face: 

FYI: tested it with both `black` and `MONAI` and there were no false positives. Test were done with py38 and py312

henzef:
No false positives in my codebase either. Thanks for the quick fix!

bugbear.py
```diff
@@ -1603,6 +1605,46 @@ def compose_call_path(node):
         yield node.id
 
 
+def _tansform_slice_to_py39(slice: ast.Slice) -> ast.Slice | ast.Name:
+    """Transform a py38 style slice to a py39 style slice.
```
cooperlees, line 1609: TIL - Thanks for this.

cooperlees (approved):
Thanks for the quick fix with great test coverage! I really appreciate great contributors like yourself.

PR #476, opened 2024-06-23
Contextvar mutable or call default

### LINKED ISSUE #475: Feature Request: expand B006 and B008 to also cover `ContextVar` defaults
```python
from contextvars import ContextVar, copy_context

cv: ContextVar[list[int]] = ContextVar("cv", default=[])

def append_to_cv_list():
    cv.get().append(0)

ctx = copy_context()
ctx.run(append_to_cv_list)
ctx.run(append_to_cv_list)
print(cv.get())
```

This will print `[0, 0]` instead of the expected `[]`, very similar to B006 and B008. Can either raise B006 & B008, or just reuse their machinery but raise a new error code.

jakkdl:
fixes #475

Not the prettiest extension of `FunctionDefDefaultsVisitor`, and left e.g. `B006.mutable_literals` as bound to `B006` and the naming of `b008_extend_immutable_calls`. I'll fix those later unless we want to reuse B006/B008 for `ContextVar`s, or maybe just a couple comments saying that B039 also uses them.

Also adding 3.13 to CI to check for any problems there

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev"]
 
         os: [ubuntu-latest]
 
--- a/README.rst
+++ b/README.rst
@@ -201,6 +201,8 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B038**: **Moved to B909** - Found a mutation of a mutable loop iterable inside the loop body. Changes to the iterable of a loop such as calls to `list.remove()` or via `del` can cause unintended bugs.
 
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13-dev"]
```
cooperlees, line 12: Happy to take it as we've lucked out. We should do this in dedicated PRs in future in case there is fallout to deal with.

cooperlees (approved):
Sounds good to me. This is a nice bug to ignore which fits in with out things we highlight. Many thanks!

jakkdl:
Added some comments to note where b006/b008 stuff also applies to b039, and fixed isort. Ready to merge unless @Zac-HD wants to have opinions

Zac-HD (approved):
Looks great - thanks, @jakkdl for implementation and @cooperlees for review 🙏

PR #477, opened 2024-06-24
add b040: exception with note added not reraised or used

### LINKED ISSUE #474: Feature request: lint for dropped exception after `.add_note()`
```python
try:
    something()
except Exception as e:
    e.add_note("some helpful context")
```

This code is easy to write, but it has a nasty problem: you've just suppressed the exception!  I think there should be a lint warning which triggers if you `.add_note()` an exception which is not either followed by a `raise` or `raise e`, or included in a function call or rhs of an assignment expression.

jakkdl:
Fixes #474 

There's some alternatives when it comes to the scope of this check, namely whether these two should be caught
##### 1
```python
def foo():
  my_exc = ValueError()
  my_exc.add_note("I want to use this variable")
  # ... but then completely forgets to raise
```
I.e. arbitrary exception created anywhere, that appears to only be used for having a note added. This one gets tricky as its complexity approaches that of a type checker, which usually handles unused variables. If we assume that only exceptions will ever have a method called `add_note` we can sidestep trying to do any type inference.
Though this would hit false alarms with stuff like
```python
def my_add_note(e: Exception) -> None:
  e.add_note("foo")
```
so it probably isn't worth it.
##### 2

```python
try:
  ...
except:
  f = ValueError()
  f.add_note("")
```

This is essentially the same as the first one, except it's happening within an `except` scope and much more clearly looks like a bug.

Handling either case will require `exceptions_tracked` to be a full dict, and the logic in a bunch of places to be beefed up a bit. If we don't care to go with that, I will simplify the type (and rename it, it's an awful name atm) of `exceptions_tracked` and simplify/clean up some code handling it.

Handling the lambda case, or other functions defined inside an except handler, wouldn't be super complicated - but I don't really think it's worth even the slight complexity.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -55,7 +55,7 @@ class BugBearChecker:
     filename = attr.ib(default="(none)")
     lines = attr.ib(default=None)
     max_line_length = attr.ib(default=79)
-    visitor = attr.ib(init=False, default=attr.Factory(lambda: BugBearVisitor))
+    visitor = attr.ib(init=False, factory=lambda: BugBearVisitor)
     options = attr.ib(default=None)
 
     def run(self):
@@ -356,17 +356,23 @@ class ExceptBaseExceptionVisitor(ast.NodeVisitor):
         return super().generic_visit(node)
 
 
+@attr.define
+class B040CaughtException:
+    name: str
+    has_note: bool
+
+
 @attr.s
 class BugBearVisitor(ast.NodeVisitor):
     filename = attr.ib()
     lines = attr.ib()
-    b008_b039_extend_immutable_calls = attr.ib(default=attr.Factory(set))
-    b902_classmethod_decorators = attr.ib(default=attr.Factory(set))
-    node_window = attr.ib(default=attr.Factory(list))
-    errors = attr.ib(default=attr.Factory(list))
-    futures = attr.ib(default=attr.Factory(set))
-    contexts = attr.ib(default=attr.Factory(list))
-    exceptions_tracked: dict[str, bool | None] = attr.ib(default=attr.Factory(dict))
+    b008_b039_extend_immutable_calls = attr.ib(factory=set)
+    b902_classmethod_decorators = attr.ib(factory=set)
+    node_window = attr.ib(factory=list)
+    errors = attr.ib(factory=list)
+    futures = attr.ib(factory=set)
+    contexts = attr.ib(factory=list)
+    b040_caught_exception: B040CaughtException | None = attr.ib(default=None)
 
     NODE_WINDOW_SIZE = 4
     _b023_seen = attr.ib(factory=set, init=False)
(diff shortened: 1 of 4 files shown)


bugbear.py
```diff
@@ -366,6 +366,7 @@ class BugBearVisitor(ast.NodeVisitor):
     errors = attr.ib(default=attr.Factory(list))
     futures = attr.ib(default=attr.Factory(set))
     contexts = attr.ib(default=attr.Factory(list))
+    exceptions_tracked: dict[str, bool | None] = attr.ib(default=attr.Factory(dict))
```
cooperlees, line 369: This hilights w should type annotate all this code. https://github.com/PyCQA/flake8-bugbear/issues/478

bugbear.py
```diff
@@ -428,11 +429,16 @@ def visit(self, node):
 
         self.check_for_b018(node)
 
-    def visit_ExceptHandler(self, node):
+    def visit_ExceptHandler(self, node):  # noqa: C901 # too complex
```
cooperlees, line 432: Wonder if it's worth creating an issue to break this up ...

cooperlees (changes requested):
Thanks!

In regards to the questions:
1) Agree, don't think it's worth it - especially with false positives
2) I think since it's in an `except` clause something should be done with the exception there .,.. so it's a bug we should care about. Up to you how far you want to go in regards to lambdas and other 
- Happy to start catching the blatant bugs - But let's note possible enhancements in the issue as someone else might want to implement.

Will request changes so you can continue on here :)

bugbear.py
jakkdl: It is quite messy, with code for B013, B029, B030, B036 and B040 interwoven. Fairly straightforward to split out B013, B029 and B030 though - so I'll just go ahead and do that.

jakkdl:
> 2. I think since it's in an `except` clause something should be done with the exception there .,.. so it's a bug we should care about.

This one is tricky if we want to avoid a false alarm for
```python
except:
  f = ValueError()
  my_exceptions.append(f)
  f.add_note("")
```

We don't want to try to detect creation of exceptions, as handling anything other than built-in exceptions would require full-on type tracking. So what we can do is detect when `add_note` is used on an object. But in this example we would need to write a custom visitor to visit twice to pick up that `f` is being used because it happens before `add_note`. I guess doing an `ast.walk` for `.add_note()` calls is a decent solution though.

And there's other thorny cases that will raise false alarms even if we limit it to "the `add_note` is in an `ExceptHandler`".
```python
my_exc = ValueError()
try:
  do_thing()
except ValueError as e:
  my_exc.add_note(str(e))
raise my_exc
```

The reason the straightforward case is not vulnerable to this kind of stuff is that the caught exception gets deleted at the end of the exception handler, so we know for sure it's limited to the scope of the `ExceptHandler`. 
So unless we're fine with the occasional false alarm I suspect I shouldn't bother trying to handle this case either. Maybe the ruff folks have sufficient infra that it's plausible for them to do so if/when they reimplement it.

cooperlees:
Hmm, yeah. Some interesting false positives here. I'm happy with not covering this and starting simpler. You do what you want here. I think it's better to start with less false positives and do dedicated PRs handling more complex edge cases in general.

You have merge conflicts now to work out now too.

jakkdl:
simplified and cleaned up the code, should be ready for final review.

cooperlees (approved):
Many thanks and very nice. Love the tests with nice comments to see what you're aiming for over the file.

PR #481, opened 2024-07-02
Do not trigger B901 with explicit Generator return type

### LINKED ISSUE #480: Suppress B901 if function explicitly returns Generator
I got a B901 warning (well, from ruff, not directly from flake8-bugbear) about a function where I use a `return` statement in a generator. But it was on purpose; returning from a generator is occasionally useful.

I propose to not trigger B901 if the function has an explicit `Generator[X, Y, Z]` type annotation, because that indicates the author explicitly thought about it.

JelleZijlstra:
Fixes #480.

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1192,9 +1192,26 @@ class BugBearVisitor(ast.NodeVisitor):
         for child in node.body:
             yield from _loop(node, child)
 
-    def check_for_b901(self, node):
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Sounds good to me. Couldn't be much more explicit.

Will prob need merge conflicts fixed.

Issue #323, opened 2022-12-16
Recommend Counter() in place of defaultdict(int)


peterjc:
I am suggesting a new check to recommend replacing ``defaultdict(int)`` with ``Counter()``, both of which are from the standard library ``collections``, due to the former having a major gotcha with memory usage.

Consider this test case:

```python
from collections import Counter
from collections import defaultdict

# Profiled using;
# $ mprof run defaultdict_leak.py && mprof plot
if True:
    # Apparent memory leak
    # Over 700MB using Python 3.9.13 on macOS
    # Over 900MB using Python 3.9.15 on Linux
    counts = defaultdict(int)
else:
    # Low and flat memory
    # about 16MB using Python 3.9.13 on macOS
    # about 19MB using Python 3.9.15 on Linux
    counts = Counter()

# Setup some sparse data
for x in [12,34,56,78,90]:
    counts[str(x)] = 123 + x

threshold = 100
excesses = 0
for x in range(int(1e7)):
    if threshold < counts[str(x)]:
        excesses += counts[str(x)]
print(f"Sum of entries exceeding the threshold {threshold} is {excesses}")
```

The apparent memory leak with ``defaultdict`` is due to the documented behaviour of recording the missing keys in the dictionary with the default value, https://docs.python.org/3/library/collections.html#collections.defaultdict says:

> ... provide a default value for the given key, this value is inserted in the dictionary for the key, and returned.
>
> ...
>
> When each key is encountered for the first time, it is not already in the mapping; so an entry is automatically created using the default_factory ...

I assume I'm not the only person to have used ``defaultdict`` without appreciating this, and in the case of ``defaultdict(int)`` there is good alternative in ``Counter()``

cooperlees:
Howdy - Good suggestion. TIL about `Counter` and I use `defaultdict(int)` a lot. With it being a dict subclass what fixes the memory? https://docs.python.org/3/library/collections.html#collections.Counter - Couldn't see anything noted here ... I guess the bug is in defaultdict itself and not dict (otherwise there would be leaks all over python as dicts are used everywhere internally).

Has the memory problem been fixed in later pythons (i.e. in >= 3.10) with defaultdict? If not, has a bug been raised? I'm torn with this one as it feels we could get it fixed in cpython, but maybe recommending it could be worth it. Would love others opinions here ...

peterjc:
[...]

PR #489, opened 2024-08-15
Added new check B910 to suggest using Counter() instead of defaultdict(int)

dejvidq:
Implementing #323.

I'm not sure if the number B910 is OK, but it's just the next available number (not counting B950).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -273,6 +273,8 @@ on the first line and urls or paths that are on their own line::
       "https://some-super-long-domain-name.com/with/some/very/long/paths"
   )
 
+**B910**: Use Counter() instead of defaultdict(int) to avoid excessive memory use as the default dict will record missing keys with the default value when accessed.
+
 
 How to enable opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/bugbear.py
+++ b/bugbear.py
@@ -503,6 +503,7 @@ class BugBearVisitor(ast.NodeVisitor):
         self.check_for_b034(node)
         self.check_for_b039(node)
         self.check_for_b905(node)
+        self.check_for_b910(node)
 
         # no need for copying, if used in nested calls it will be set to None
         current_b040_caught_exception = self.b040_caught_exception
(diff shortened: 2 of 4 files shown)


README.rst
```diff
@@ -273,6 +273,8 @@ on the first line and urls or paths that are on their own line::
       "https://some-super-long-domain-name.com/with/some/very/long/paths"
   )
 
+**B910**: Use Counter() instead of defaultdict(int) to avoid memory leaks. Using the latter can result in using as much as ~1000 MB of memory even for small and simple dicts. Refactoring the code to use the former can reduce memory usage from mentioned ~1000 MB to flat ~20 MB.
```
JelleZijlstra, line 276: Where does this number come from? I would expect it depends on the size of the data; i.e., you can construct a Counter that is a 1000 MB and also a defaultdict that is 20 MB.

It appears the difference is that when you access a key that does not exist, Counter simply returns 0, but defaultdict sets the value to 0 in the dict as well. How much that affects your memory usage obviously depends on how often you access keys that don't exist. We should mention the pattern that leads to the difference, not the exact number, because the exact number is going to depend on your use case.

peterjc: I think the numbers were based on the test case I provided in #323, and yes they will depend on the number of missing keys accessed.

How about: "Use Counter() instead of defaultdict(int) to avoid excessive memory use as the default dict will record missing keys with the default value when accessed."

bugbear.py
```diff
@@ -2380,6 +2391,9 @@ def visit_Lambda(self, node) -> None:
         "B909 editing a loop's mutable iterable often leads to unexpected results/bugs"
     )
 )
+B910 = Error(
+    message="B910 use Counter() instead of defaultdict(int) to avoid memory leaks"
```
peterjc, line 2395: I'd change "memory leaks" to "memory usage", although that is perhaps a matter of perspective.

README.rst
JelleZijlstra: Yes, that sounds good.

cooperlees: +1 - Thanks all for discussing!

dejvidq: Thanks, I agree that it wasn't the best decision to mention any numbers there. I changed that.

bugbear.py
dejvidq: Right. Fixed.

README.rst
```diff
@@ -273,6 +273,8 @@ on the first line and urls or paths that are on their own line::
       "https://some-super-long-domain-name.com/with/some/very/long/paths"
   )
 
+**B910**: Use Counter() instead of defaultdict(int) to avoid excessive memory use as the default dict will record missing keys with the default value when accessed."
+
```
peterjc, line 277: Looks like an unwanted trailing double quote?

dejvidq: Yes 😕 
Fixed

cooperlees (approved):
Thanks for eventually implementing it with tests. Appreciate it.

PR #486, opened 2024-08-05
Replace hash with unsafe_hash

kasium:
Background: https://www.attrs.org/en/stable/changelog.html#deprecations

Closes #485

Issue #485, opened 2024-08-05
DeprecationWarning: The `hash` argument is deprecated in favor of `unsafe_hash`


kasium:
When running some flake8 tests also including flake8-bugbear I can see the following DeprecationWarning

```
.nox/test_flake8-3-10-flake8-7/lib/python3.10/site-packages/bugbear.py:49: in <module>
    @attr.s(hash=False)
.nox/test_flake8-3-10-flake8-7/lib/python3.10/site-packages/attr/_make.py:1291: in attrs
    warnings.warn(
E   DeprecationWarning: The `hash` argument is deprecated in favor of `unsafe_hash` and will be removed in or after August 2025.
```

cooperlees:
Thanks for the report. PR welcome.

PR #490, opened 2024-08-21
Bump required attrs version

henzef:
Due to #485 / #486, bugbear now requires at least attrs 22.2.0


cooperlees (approved):
O, ok, many thanks!

Hope this isn't biting to many people.

PR #491, opened 2024-09-05
README: Add anchors for rules and config options

jtojnar:
This will allow us to link to them directly.

Using the syntax from https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-a-custom-anchor

Example: https://github.com/jtojnar/flake8-bugbear/tree/readme-anchors#user-content-b950

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,24 +58,34 @@ If you'd like to do a PR we have development instructions `here <https://github.
 List of warnings
 ----------------
 
+.. _B001:
+
 **B001**: Do not use bare ``except:``, it also catches unexpected events
 like memory errors, interrupts, system exit, and so on.  Prefer ``except
 Exception:``.  If you're sure what you're doing, be explicit and write
 ``except BaseException:``.  Disable ``E722`` to avoid duplicate warnings.
 
+.. _B002:
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -58,76 +58,110 @@ If you'd like to do a PR we have development instructions `here <https://github.
 List of warnings
 ----------------
 
+.. _B001:
+
 **B001**: Do not use bare ``except:``, it also catches unexpected events
```
jtojnar, line 63: We may also want to add a link to name so that users can obtain it without having to dig through the source code.

cooperlees (changes requested):
Thanks for this. This would be useful, I can see.

Is there any way we can make rstcheck happy here?
- Seems it's a known problem: https://github.com/rstcheck/rstcheck-core/issues/4

Can we look at adding the correct `--ignore-directives` to the [pre-commit config](https://github.com/PyCQA/flake8-bugbear/blob/main/.pre-commit-config.yaml) please to not error rstcheck since this is a known problem.

cooperlees (approved):
Going to merge and fix the linting locally

PR #496, opened 2024-10-30
Add B041: Duplicate keys in dictionary literals

yangdanny97:
This is the counterpart to the duplicate dictionary key linter in pyflakes: https://github.com/PyCQA/pyflakes/pull/72

Pyflakes only emits an error if the values are different for the duplicate keys. This lint copies most of the logic but does the opposite - when we have multiple identical keys with the identical values, we emit an error on all but the first entry.

As with Pyflakes, it only works when the keys are tuples, literals, or variable names.

This could have been _much_ simpler if we could just emit an error on all duplicate keys regardless of their values, but the conversation in the PR for B033 indicated that we should avoid duplication with Pyflakes.

https://github.com/PyCQA/flake8-bugbear/pull/373

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -205,6 +205,8 @@ second usage. Save the result to a list if the result is needed multiple times.
 
 **B040**: Caught exception with call to ``add_note`` not used. Did you forget to ``raise`` it?
 
+**B041**: Repeated key in dictionary literal. The last value will override any previous values.
+
 Opinionated warnings
 ~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


amyreese:
Would it be better to build this as a generic "duplicate key" rule, regardless of whether the content is the same, and just advise users to disable any relevant flake8/pyflakes rules to avoid overlapping violations?

yangdanny97:
The other branch of my fork has a simplified version: https://github.com/yangdanny97/flake8-bugbear/blob/main/bugbear.py#L640

But as-implemented it only supports Constant keys, so would have less coverage in some areas compared to pyflakes. If we wanted it to be a superset of the errors caught by pyflakes, we would still need to pull over some of the `convert_to_value` logic

cooperlees (approved):
I think this covers enough to start. We could enhance it / tune it moving forward.

Tests are pretty good coverage too ...

I will add one more scenence of info into the README tho.

Issue #419, opened 2023-10-28
Improve unit test runner


JelleZijlstra:
This project's unit tests are sort of annoying to work with as you have to manually type out line numbers in the test file. This is especially bad if we reformat something and a bunch of line numbers change.

It would be nice to switch to flake8-pyi's setup or a variant (https://github.com/PyCQA/flake8-pyi/blob/main/tests/test_pyi_files.py), which puts the expected errors inline in the test files.

AlexWaygood:
Looks like bugbear currently asserts the column number an error occurs on, as well as the line number — you won't be able to do that with a flake8-pyi-esque setup.

I don't think that's a huge issue, though. Having also contributed to both projects, I agree that flake8-pyi's test suite is _much_ nicer to work with! This sounds like a great idea.

cooperlees:
FWIW - We've at least disabled black on test code so we don't reformat and change numbers as much with formatting changes.

PR #500, opened 2024-12-05
add except* support to B012&B025

jakkdl:
This PR creates a `visit_TryStar` that mostly just calls `visit_Try`, enabling B012 & B025 to work with `try/except*`

Because `ExceptHandler` is also used for `TryStar`, we already have support for free for B014, B029, B030, B036 and B904.
I opted to add tests for them anyway, but it did end up with a somewhat silly amount of added tests - most of which are just pure copy-paste's of their original test files. Another option is writing a single `except_star.py` file that tests all of them, or writing a transformer test that goes through the original files, does a search+replace for `except`->`except*`, and then tests them on the fly. Though the latter option would also involve some rejigging to get the expected errors shared from the original `test_bxxx` functions.
Another case for #419 

B001 is the only except-related rule I found that's not revelant, since `except*:` is a syntax error.

EDIT: error message updates added in https://github.com/PyCQA/flake8-bugbear/pull/500/commits/adca610ce7d1d50c856586d6cb8465fe37dff10b
~~I have *not* updated error messages. For B012 (unclear if necessary) and B025 it'd be fairly straightforward, could make `visit_TryStar` pass a parameter to `check_for_b012`/`check_for_b025`. For the others it'd have to involve a property `BugBearVisitor.in_trystar`... I guess I should just do that, even if somewhat tedious.~~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -360,6 +360,11 @@ Change Log
 ----------
 
 
+FUTURE
+~~~~~~
+
+* B012 and B025 now also handle try/except*
+
 24.10.31
 ~~~~~~~~
 
--- a/bugbear.py
+++ b/bugbear.py
@@ -243,7 +243,7 @@ def _flatten_excepthandler(node: ast.expr | None) -> Iterator[ast.expr | None]:
         yield expr
 
 
-def _check_redundant_excepthandlers(names: Sequence[str], node):
+def _check_redundant_excepthandlers(names: Sequence[str], node, in_trystar):
     # See if any of the given exception names could be removed, e.g. from:
     #    (MyError, MyError)  # duplicate names
     #    (MyError, BaseException)  # everything derives from the Base
@@ -275,7 +275,7 @@ def _check_redundant_excepthandlers(names: Sequence[str], node):
         return B014(
             node.lineno,
             node.col_offset,
-            vars=(", ".join(names), as_, desc),
+            vars=(", ".join(names), as_, desc, in_trystar),
         )
     return None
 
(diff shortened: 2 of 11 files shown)


cooperlees:
I think the simple copy paste for tests is usually worth it over more complext helper functions etc. - But I see here how our tests simplicity create annoying amounts of fun ...

I don't think I want a except_star.py ... As for updated messages, it's just the same error inside a try/except right? If so, I don't think we need to change the messages. If I've missed context just share. But PR seems good to me.

jakkdl:
> I think the simple copy paste for tests is usually worth it over more complext helper functions etc. - But I see here how our tests simplicity create annoying amounts of fun ...
> I don't think I want a except_star.py ... 

I might give #419 a shot, in which case keeping to the current structure also makes it simpler. The downside of copy-paste would be having to maintain multiple copies, but that shouldn't be much of an issue here.

> As for updated messages, it's just the same error inside a try/except right? If so, I don't think we need to change the messages. If I've missed context just share. But PR seems good to me.

Yeah it's literally just saying `Using except* (): with an empty tuple does not handle/catch anything...` instead of `Using except (): with an empty tuple does not handle/catch anything...`. I can revert the last commit if you think the complexity cost to the code is higher than the gain from printing the `*`. Given that you can't have multiple `except/except*` on the same line it's unlikely to lead to confusion.

cooperlees (approved):
All sounds good to me. I'll leave the complexity to you.

One thing I don't see how it works, are our checks for `except*` now somehow recursive to now include all subclasses for the specified exception(s)? If so how, is it inbuilt to the methods used somehow?

jakkdl:
> All sounds good to me. I'll leave the complexity choice to you.

I'll just keep it as is then. Matches ruff as well.

> One thing I don't see how it works, are our checks for `except*` now somehow recursive to now include all subclasses for the specified exception(s)? If so how, is it inbuilt to the methods used somehow?

I don't follow, what is this referring to?

cooperlees:
I think I'm mis-understanding the use of `except*` ... I've never used it. I'll trust you here it's all good.

PR #501, opened 2024-12-07
Avoid raising `B028` if `warnings.warn` contains `*args` or `**kwargs`

harupy:


tests/b028.py
```diff
@@ -11,3 +11,7 @@
 warnings.warn("test", DeprecationWarning, stacklevel=1)
 warnings.warn("test", DeprecationWarning, 1)
 warnings.warn("test", category=DeprecationWarning, stacklevel=1)
+args = ("test", DeprecationWarning, 1)
+warnings.warn(*args)
+kwargs = {"message": "test", "category": DeprecationWarning, "stacklevel": 1}
+warnings.warn(**kwargs)
```
cooperlees, line 17:
```suggestion
warnings.warn(*args)
kwargs = {"message": "test", "category": DeprecationWarning, "stacklevel": 1}
warnings.warn(**kwargs)
warnings.warn(*args, **kwargs)
```
Do we need to handle both? If so, let's explicitly test.

cooperlees (approved):
Thanks! LGTM - Makes sense to down noise when we can't be sure.

Just one question on maybe a little more test coverage.

harupy:
@cooperlees Applied your suggestion! Could you take another look?

cooperlees:
Many thanks for your contribution! Will try get a release out this week.

PR #502, opened 2024-12-07
Add B911: itertools.batched without strict=

### LINKED ISSUE #498: Rule proposal: require explicit `strict=` argument for `itertools.batched`
[As of Python 3.13](https://docs.python.org/3/library/itertools.html#itertools.batched), `itertools.batched` has a `strict` parameter that defaults to `False`.  By default, the batches might not be of the same size, which may cause subtle bugs. When`strict=True`, it raises `ValueError` if the final batch is not the same size as the rest.

This seems analogous to the existing rule `B905` which requires an explicit `strict=` parameter for `zip`. Hence, I think it makes sense if there is a similar rule for `itertools.batched`.

I volunteer to implement the rule, if accepted!

tjkuson:
Checks calls to `itertools.batched` and `batch` similar to the implementation of B031.

Closes #498

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -258,6 +258,10 @@ This is meant to be enabled by developers writing visitors using the ``ast`` mod
 
 **B910**: Use Counter() instead of defaultdict(int) to avoid excessive memory use as the default dict will record missing keys with the default value when accessed.
 
+**B911**: ``itertools.batched()`` without an explicit `strict=` parameter set. ``strict=True`` causes the resulting iterator to raise a ``ValueError`` if the final batch is shorter than ``n``.
+
+The ``strict=`` argument was added in Python 3.13, so don't enable this flag for code that should work on <3.13.
+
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -258,6 +258,10 @@ This is meant to be enabled by developers writing visitors using the ``ast`` mod
 
 **B910**: Use Counter() instead of defaultdict(int) to avoid excessive memory use as the default dict will record missing keys with the default value when accessed.
 
+**B911**: ``itertools.batched()`` without an explicit `strict=` parameter set. ``strict=True`` causes the resulting iterator to raise a ``ValueError`` if the final batch is shorter than ``n``.
```
cooperlees, line 261: Nice, you get the cool Porsche number.

cooperlees (approved):
LGTM - Thanks for the simple implementation + good tests.

PR #503, opened 2025-01-13
Ignore B028 if skip_file_prefixes is used

### LINKED ISSUE #497: Ignore B028 if skip_file_prefixes is used
If a warnings.warn call contains no stacklevel but skip_file_prefixes is set, no issue should be raised, because if skip_file_prefixes is used, the author basically skips stack entries and also stacklevel has an implicit miniumum value of 2 automatically

kasium:
Closes #497


cooperlees (approved):
LGTM - I'll add into the changes rather than block on that :)

kasium:
Ah sorry. Forgot it. Thanks for adding them

PR #507, opened 2025-03-30
Move project to >= 3.9

cooperlees:
- CI failed as we'd hoped and it's time to move to >= 3.9 cpython to match flake8
- Here we update all configuration and metadata etc. etc.

This fixes "flake8_py_version_check" CI failing like: https://github.com/PyCQA/flake8-bugbear/actions/runs/14150950748/job/39644039034

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -475,7 +475,7 @@ Change Log
 UNRELEASED
 ~~~~~~~~~~
 
-* flake8-bugbear is now >= 3.9 project like flake8>=7.2.0
+* flake8-bugbear now requires at least Python 3.9, like flake8>=7.2.0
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -475,6 +475,7 @@ Change Log
 UNRELEASED
 ~~~~~~~~~~
 
+* flake8-bugbear is now >= 3.9 project like flake8>=7.2.0
```
JelleZijlstra, line 478:
```suggestion
* flake8-bugbear now requires at least Python 3.9, like flake8>=7.2.0
```

JelleZijlstra (approved):

PR #509, opened 2025-05-09
drop 3.8, unpin mypy dep in tox, fix typing errors, add some type hints

jakkdl:
the `partial` approach didn't pass typing anymore, though I didn't dig terribly deep into why that might be the case, I simply replaced it with a factory class.

testing 3.14 is blocked by hypothesmith->libcst->PyO3 https://github.com/PyO3/pyo3/issues/5107

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -10,9 +10,9 @@ import sys
 import warnings
 from collections import Counter, defaultdict, namedtuple
 from contextlib import suppress
-from functools import lru_cache, partial
+from functools import lru_cache
 from keyword import iskeyword
-from typing import Dict, Iterable, Iterator, List, Sequence, Set, Union
+from typing import Dict, Iterable, Iterator, List, Sequence, Set, Union, cast
 
 import attr
 import pycodestyle  # type: ignore[import-untyped]
@@ -58,7 +58,7 @@ class BugBearChecker:
     visitor = attr.ib(init=False, factory=lambda: BugBearVisitor)
     options = attr.ib(default=None)
 
-    def run(self):
+    def run(self) -> Iterable[tuple[int, int, str, type]]:
         if not self.tree or not self.lines:
             self.load_file()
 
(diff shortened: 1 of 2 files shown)


bugbear.py
```diff
@@ -243,7 +243,9 @@ def _flatten_excepthandler(node: ast.expr | None) -> Iterator[ast.expr | None]:
         yield expr
 
 
-def _check_redundant_excepthandlers(names: Sequence[str], node, in_trystar):
+def _check_redundant_excepthandlers(
+    names: Sequence[str], node: ast.ExceptHandler, in_trystar: str
```
JelleZijlstra, line 247: surely this?
```suggestion
    names: Sequence[str], node: ast.ExceptHandler, in_trystar: bool
```

jakkdl: nope, I thought so too at first but it's from
https://github.com/PyCQA/flake8-bugbear/blob/826c071ce81b70562daa2cc7ba37cef1d5d74a05/bugbear.py#L391-L392

bugbear.py
```diff
@@ -477,10 +479,10 @@ def visit_ExceptHandler(self, node: ast.ExceptHandler) -> None:
             self.errors.append(B040(node.lineno, node.col_offset))
         self.b040_caught_exception = old_b040_caught_exception
 
-    def visit_UAdd(self, node) -> None:
+    def visit_UAdd(self, node: ast.UAdd) -> None:
         trailing_nodes = list(map(type, self.node_window[-4:]))
         if trailing_nodes == [ast.UnaryOp, ast.UAdd, ast.UnaryOp, ast.UAdd]:
-            originator = self.node_window[-4]
+            originator = cast("ast.UnaryOp", self.node_window[-4])
```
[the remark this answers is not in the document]
JelleZijlstra: It looks worse

JelleZijlstra (approved):

PR #511, opened 2025-05-22
revamp testing infra

### LINKED ISSUE #419: Improve unit test runner

jakkdl:
fixes #419 :rocket: 

The general approach is copied from the test runner I wrote for https://github.com/python-trio/flake8-async, minus a bunch of bells and whistles that aren't necessary for an initial implementation.

I had to move the error classes to be defined inside a dict in order for `isort` to be happy, as otherwise we'd be importing a ton of objects that were never used other than indirectly through `eval`. I also created an `add_error` helper method to simplify the code.

Aside from those the changes to the code are minimal, the vast majority of the changes in the admittedly *huge* diff is simply moving the expected result to be inline in the test files.

The B950 tests were parametrized in a way that wasn't trivial to translate to this system, the column changes are due to line lengths being different with inlined error data.

Credits to Claude Code that parsed the test data and crafted sed commands to update the test files, but I kept a close eyes on all the changes so there should be zero impact on actual functionality.

### DIFF
--- a/tests/test_bugbear.py
+++ b/tests/test_bugbear.py
@@ -80,7 +80,7 @@ def _parse_eval_file(test: str, content: str) -> tuple[list[error], Namespace | 
         if "#" not in line:
             continue
 
-        # get text between `error:` and (end of line or another comment)
+        # get text between `B\d\d\d:` and (end of line or another comment)
         k = re.findall(r"(B\d\d\d):([^#]*)(?=#|$)", line)
         for err_code, err_args in k:
             # evaluate the arguments as if in a tuple


cooperlees (approved):
O wow, nice! Thanks!

This mainly gets us away from hard coding line numbers right?

jakkdl:
> O wow, nice! Thanks!
> 
> This mainly gets us away from hard coding line numbers right?

yes! You instead add a magic comment to the line you expect to have an error. I should probably add some instructions for it in `DEVELOPMENT.md`

cooperlees:
> I should probably add some instructions for it in DEVELOPMENT.md

Please! Than I'll merge. This is awesome ....

jakkdl:
> > I should probably add some instructions for it in DEVELOPMENT.md
> 
> Please! Then I'll merge. This is awesome ....

done! Also moved all the test files into a subdirectory. Idk if fully necessary, but makes `test_bugbear.py` easier to find and avoids having to specify a bunch of regexes and requiring eval_files to be named `b.*`

DEVELOPMENT.md
```diff
@@ -42,23 +42,48 @@ cd flake8-bugbear
...
+
+The expected errors are specified by adding comments on the line where the error is expected, using the format `# <error_code>: <col_offset>[, <var1>][, <var2>][...]`. E.g.
+```python
+x = ++n  # B002: 4
+try:
+    ...
+except* (ValueError,):  # B013: 0, "ValueError", "*"
+    ...
+```
+The error code should be in the `error_codes` dict, and the other values are passed to `eval` so should be valid python objects.
```
JelleZijlstra, line 57: I don't understand what the var1, var2, ... do, could you clarify?

> the other values are passed to `eval` so should be valid python objects

The example uses `"*"` which fails eval.

jakkdl: Tried to clarify it, is it better now?

> The example uses "*" which fails eval.

well, `eval('"*"')` is fully valid. But I tried making that more explicit as well.

Feel free to suggest any changes if it still doesn't make sense, the relevant logic is

https://github.com/PyCQA/flake8-bugbear/blob/f7d5123f7aaefe27b9a1e8bbdf76b26157dfc336/tests/test_bugbear.py#L83-L92

PR #512, opened 2025-05-26
add B042: exception class with no, or malformed, super().__init__()

### LINKED ISSUE #421: Feature request: a rule for calling super().__init__() in custom exception's __init__()
```python
class CustomException(Exception):
    def __init__(self, custom: int, arguments: str) -> None:
        self.msg = f"Something specific happened with custom: {int} and {arguments}"
```

Above exception class has an issue, in my personal experience it's quite common. The implementer assumes, that they will use the `self.msg` parameter during exception handling. The issue araises when the exception gets anywhere else as the its string representation is as follows:

```python
>>> raise CustomException(42, 'test')
  File "<stdin>", line 1, in <module>
__main__.CustomException: (42, 'test')
```

and when the exception is instantiated with keyword arguments, we lose the information about the arguments' values.

```python
>>> raise CustomException(custom=1, arguments='two')
  File "<stdin>", line 1, in <module>
__main__.CustomException
```

How to fix it? By calling `super().__init__()` in the constructor. That way the string representation of the custom exception will contain the error message.

```python
class CustomException(Exception):
    def __init__(self, custom: int, arguments: str) -> None:
        self.msg = "Something specific happened with custom: {int} and {arguments}"
        super().__init__(self.msg)
>>> raise CustomException(1, 'two')
  File "<stdin>", line 1, in <module>
__main__.CustomException: Something specific happened with custom: 1 and two
>>> raise CustomException(custom=1, parameters='two')
  File "<stdin>", line 1, in <module>
__main__.CustomException: Something specific happened with custom: 1 and two
```

**Specification proposal**

If a class:
* inherits from from `Exception`,
* doesn't implement own `__str__()` method,
* its `__init__()` method doesn't enforce any positional-only argument, 
* and doesn't call `super().__init__()` in its `__init__()` method

Then:
* new rule should raise an error, calling to fix it.

jakkdl:
fixes #421

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -1735,8 +1735,8 @@ class BugBearVisitor(ast.NodeVisitor):
             for base in node.bases:
                 if isinstance(base, ast.Name) and is_exception(base.id):
                     break
-                else:
-                    return
+            else:
+                return
 
         # iterate body nodes looking for __init__
         for fun in node.body:
--- a/bugbear.py
+++ b/bugbear.py
@@ -613,6 +613,7 @@ class BugBearVisitor(ast.NodeVisitor):
         self.check_for_b903(node)
         self.check_for_b021(node)
         self.check_for_b024_and_b027(node)
+        self.check_for_b042(node)
         self.generic_visit(node)
 
     def visit_Try(self, node) -> None:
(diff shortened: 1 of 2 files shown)


cooperlees (approved):
Sorry I took so long, We need a readme update but I'll add since I took so long here to review.

cooperlees:
https://github.com/PyCQA/flake8-bugbear/commit/5962bbd925aebae15c1ea5cf58d347a33c74265c - Added to README

jakkdl:
no worries, thanks! 🚀

PR #527, opened 2025-11-05
#525 disable B042

shg-rtx:
As discussed in #525 until better documentation and fix for false positives are solved for B042, it will be disabled.

Sorry if this is not the correct way of disabling things in Flake8, I have no experience in flake8 development.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [flake8]
 # Keep in sync with .flake8. This copy here is needed for source packages
 # to be able to pass tests without failing selfclean check.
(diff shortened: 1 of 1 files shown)


cooperlees (changes requested):
This is what you'd do on your project side itself to disable the check.

This will only disable it for the flake8-bugbear project itself, we need to change code to make it not enable by default for users by default.

shg-rtx:
@cooperlees  So there is NO nice proper way of just deactivating it without making it? I actually have to go in an uncomment `self.check_for_b042?`?

PR #535, opened 2025-12-17
Add B044: `assert <generator_expression>`

### LINKED ISSUE #534: Proposed warning: `assert <generator_expression>`
For example:

```python
assert (result[i] == expected[i] for i in range(len(expected)))
assert (result[key] == expected[key] for key in result.keys() | expected.keys())
```

bugbear could warn and suggest adding `all()`.

Note that even an empty generator expression is truey:

```python
assert (False for _ in range(0))
```

so it's fine that bugbear doesn't statically know the length of the generator.

There are a couple of similar warnings:

*   F631 is a similar rule for `assert <tuple>`.
*   Proposal (2) in #440 is a similar rule for `assert <constant>`.

gschaffner:
Closes #534.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -304,6 +304,10 @@ If you define `__str__/__reduce__` in super classes this check is unable to dete
 **B043**: Do not call ``delattr(x, 'attr')``, instead use ``del x.attr``. 
 There is no additional safety in using ``delattr`` if you know the attribute name ahead of time.
 
+.. _B044:
+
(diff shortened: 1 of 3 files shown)


cooperlees:
@copilot Can you incorporate the more test cases for edge cases and apply Jelle's suggestion

cooperlees:
Copilot doesn't want to help - Going to merge and repopen the issues saying we can improve this

PR #553, opened 2026-05-13
fix(b912): only emit on Python 3.14+ where `map()` supports `strict=` (#548)

SAY-5:
Closes #548. You said "I would take a PR to add this check and let checks use it to guage to be a noop or not." Gates `check_for_b912` on `sys.version_info >= (3, 14)` so the rule no-ops on older interpreters where `map(strict=...)` would be a TypeError.


cooperlees (approved):
Nice, even with tests - Thanks!

Issue #545, opened 2026-02-25
Proposed rule: match zip() arguments and unpacked variables


jenstroeger:
In an older code blob of mine I came across this oddity:
```python
for a, b, _ in zip(iter_a, iter_b, iter_c):
    ...
```
which should be simplified to
```python
for a, b in zip(iter_a, iter_b):
    ...
```
by removing `iter_c` entirely. Turns out I had this pattern in two other places as well 🤦🏻‍♂️

Would this not be a useful idiom to look for and warn about? (Assuming `_` isn’t used in the body of the loop…)

cooperlees:
Yeah, I feel this is a unneeded code smell. So I'll accept a new lint here recommending just removing from the `zip()` rather than ignoring the element. Please add tests like all other lints and to README.

I think using "`_`" is a thing all devs should not do ... so this helps that ...

PR #556, opened 2026-05-26
AI-generated possible solution for #545 (B)

hugnia:
Hi,

We're a research team at JetBrains studying how maintainers evaluate AI-generated code.

We opened two pull requests as possible solutions to the corresponding issue, each taking a different approach. Both were generated by AI and have not yet been manually reviewed by a human. We’d like to get your perspective on these two solutions via a 10-minute survey. Regardless of your role, we want to know which solution actually saves you work, and which qualities of AI-generated code matter most when working with the codebase.

If you're open to participating in our study, just reply "I'm interested" here, and we'll send the details to the contact specified in your GitHub profile. If you'd rather not take part, feel free to close the PRs – we won’t post them again.

Thanks!

Issue #545

### DIFF
--- a/bugbear.py
+++ b/bugbear.py
@@ -593,6 +593,7 @@ class BugBearVisitor(ast.NodeVisitor):
         self.check_for_b023(node)
         self.check_for_b031(node)
         self.check_for_b909(node)
+        self.check_for_b913(node)
         self.generic_visit(node)
 
     def visit_AsyncFor(self, node: ast.AsyncFor) -> None:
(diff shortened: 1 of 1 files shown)


cooperlees:
I generally get AI to review, then sanity check code. Here I'd love some unittests please showing the check works and finds lint errors with all examples of everything it can fire on please. I think I prefer this being an optional check, so let's for with this PR.

Please also address copilot's findings.

hugnia:
Thank you so much for taking the time to review this PR and share your feedback — we genuinely appreciate it!

I should have made this clearer upfront: as mentioned in our initial message, both PRs were submitted as part of a research study on how maintainers evaluate AI-generated code. They are not intended for actual merging, and we will not be able to refine or update them. We are sorry for any confusion this caused — your time is valuable. If you would like to participate in the study and share your thoughts on the two solutions, we would love to hear from you. Otherwise, please feel free to close the PRs.

cooperlees:
Yeah, that was not clear. Please improve that for others as we as maintainers have limited time and you've now taken mine for no advantage to this project. That makes me sad.

PR #557, opened 2026-06-15
fix B031 false positive with if-else branches

wali-reheman:
when the groupby variable is used inside an if-else, only one branch executes at runtime, so using it in both branches shouldn't trigger B031.


cooperlees (changes requested):
Hi - many thanks for the potential bug fix. Can you add some example code into B031's tests to show it now working and the code so I can try understand the bug more please.

Also, please make sure pre-commit passes and let's see if copilot finds anything.

Thanks again.

Issue #465, opened 2024-03-24
B031 does not take into account if-else statements


amyasnikov:
Bugbear reports false-positive "more that one" usage of the generator even if it occurs inside 2 different branches of the if-else statement.

This code
```
import itertools

for _, group in itertools.groupby([]):
    if True:
        print(group, True)
    else:
        print(group, False)
```

produces
```
b031.py:7:15: B031 Using the generator returned from `itertools.groupby()` more than once will do nothing on the second usage. Save the result to a list, if the result is needed multiple times.
```
but it's obvious that `group` cannot be used twice here.

flake8 --version:
```
7.0.0 (flake8-bugbear: 24.2.6, mccabe: 0.7.0, pycodestyle: 2.11.1, pyflakes: 3.2.0) CPython 3.11.2 on Linux
```

cooperlees:
Thanks for reporting. Both are things we could tighten. PRs welcome.

Sonike:
I reproduced this on the current `main` branch and would like to work on the if/else false positive. I plan to make B031 count usages per mutually exclusive control-flow path, while preserving warnings for repeated uses within the same branch and for uses before or after the conditional. I'll add focused eval fixtures and run the full test/lint suite.

I'll keep this change scoped to the if/elif/else behavior; the annotation case mentioned in the comments has already been addressed separately.

PR #558, opened 2026-06-16
B031: don't count store-context references as generator uses

StressTestor:
## what

fixes a B031 false positive: type-annotating (or assigning to) the `groupby` loop variable was counted as a "use", so a single real use plus an annotation tripped "used more than once".

```python
for _, group in groupby(items):
    group: SomeType        # annotation, not a use
    do_something(group)    # the only real use
```

before: B031 fires. after: clean.

## why

the B031 usage-counter walked the loop body and counted every `Name` matching the group variable regardless of context. an annotation target (`group: T`) or an assignment target is a `Store`-context `Name`, not a use of the generator, so it inflated the count. count only `Load`-context references.

## scope

this is the type-annotation / store-context case raised in the comments on #465. the original if/else case is handled separately by #557, and the two are orthogonal: if/else uses are both loads, untouched by this change, so they don't collide. real double-uses still fire B031.

## tests

a case in `tests/eval_files/b031.py` that annotates the loop variable then uses it once, expecting no B031. it fails before the fix and passes after. full suite stays green.

Addresses #465.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -499,6 +499,7 @@ UNRELEASED
 ~~~~~~~~~~
 
 * B018: handle also useless calls such as `isinstance(x, int)` without assigning or using the result
+* B031: don't count a store-context reference (e.g. an annotation target like `group: T`) as a use of the `groupby` generator (#465)
 
 25.11.29
 ~~~~~~~~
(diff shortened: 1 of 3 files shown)


cooperlees (approved):
Nice - Thanks for the test showing it do as you claim too!

Let's just see if copilot nit picks anything useful before merge.

PR #562, opened 2026-06-25
Fix B020 false positive for attribute iterables

### LINKED ISSUE #521: B020 false positives with iterable attribute acess
Both of these loops report that the iteration variable is being reassigned, which is it clearly not.

Both are a bit complex, but it is each to see that `smoother` has attribted accessed but not assigned to.

```
    for smoother in smoother.smoothers:
        train_basis = smoother.basis[train_index]
        train_der_basis = smoother.der_basis[train_index]
        train_der2_basis = smoother.der2_basis[train_index]
        train_cov_der2 = smoother.cov_der2
        # TODO: Double check this part. cov_der2 is calculated with all data
        train_x = smoother.x[train_index]

        train_smoothers.append(
            UnivariateGenericSmoother(
                train_x, train_basis, train_der_basis, train_der2_basis,
                train_cov_der2, smoother.variable_name + ' train'))

        test_basis = smoother.basis[test_index]
        test_der_basis = smoother.der_basis[test_index]
        test_cov_der2 = smoother.cov_der2
        # TODO: Double check this part. cov_der2 is calculated with all data
        test_x = smoother.x[test_index]

        test_smoothers.append(
            UnivariateGenericSmoother(
                test_x, test_basis, test_der_basis, train_der2_basis,
                test_cov_der2, smoother.variable_name + ' test'))
```

Here `ax` and `series` have attribute access but no assignment. 

```
        for i, (ax, (series, def_name)) in enumerate(zip(axs, series)):
            if def_name != "residual":
                ax.plot(series)
            else:
                ax.plot(series, marker="o", linestyle="none")
                ax.plot(xlim, (0, 0), color="#000000", zorder=-3)
            name = getattr(series, "name", def_name)
            if def_name != "Observed":
                name = name.capitalize()
            title = ax.set_title if i == 0 and observed else ax.set_ylabel
            title(name)
            ax.set_xlim(xlim)
```

binggao1230:
Fixes #521.

## Summary
- stop B020 from treating the base name of a plain attribute iterable like `smoother.smoothers` as the iterable being reassigned
- keep method-call iterable cases such as `values.items()` covered
- add a regression case for attribute access on the loop target name

## Tests
- `uv run --python 3.13 --with '.[dev]' pytest tests/test_bugbear.py -k B020 -q`
- `uv run --python 3.13 --with '.[dev]' pytest tests/test_bugbear.py -q`
- `uv run --python 3.13 --with flake8 --with attrs flake8 bugbear.py tests/test_bugbear.py`
- `uv run --python 3.13 --with mypy --with types-flake8 --with attrs --with flake8 mypy bugbear.py`
- `git diff --check`

### DIFF
--- a/tests/eval_files/b020.py
+++ b/tests/eval_files/b020.py
@@ -38,3 +38,8 @@ for vars in [i for i in vars]:  # B020: 4, "vars"
 
 for var in sorted(range(10), key=lambda var: var.real):
     print(var)
+
+# Attribute access on the loop target name is not iterating over the same
+# object, so it should not trigger B020.
+for smoother in smoother.smoothers:
+    print(smoother)
--- a/bugbear.py
+++ b/bugbear.py
@@ -2214,6 +2214,18 @@ class FunctionDefDefaultsVisitor(ast.NodeVisitor):
 class B020NameFinder(NameFinder):
     """Ignore names defined within the local scope of a comprehension."""
 
+    def visit_Attribute(self, node) -> None:  # noqa: B906
+        pass
+
+    def visit_Call(self, node) -> None:
+        if isinstance(node.func, ast.Attribute):
+            self.visit(node.func.value)
+            self.visit(node.args)
+            self.visit(node.keywords)
+            return
+
+        self.generic_visit(node)
+
     def visit_GeneratorExp(self, node) -> None:
         self.visit(node.generators)
 


tests/eval_files/b020.py
```diff
@@ -38,3 +38,8 @@
 
 for var in sorted(range(10), key=lambda var: var.real):
     print(var)
+
+# Attribute access on the loop target name is not iterating over the same
+# object, so it should not trigger B020.
+for smoother in smoother.smoothers:
+    print(smoother)
```
cooperlees, line 45:
```suggestion
for smoother_loop in smoother.smoothers:
    print(smoother_loop)
smoother = smother_loop
```
I argue not doing this is a good practice even tho it's technically not overwriting what your iterating ...

I am a fan of being explicit even at the cost of more assignments etc.

cooperlees (commented):
I'm not sure we want to fix this, open to other perspectives tho ... and I'll go with majority ...

binggao1230:
Thanks for looking. My reasoning for treating this as a false positive is that B020 is most useful when the loop target overwrites the iterable object itself, e.g. the warning protects `for value in values:` and receiver-style calls like `for key in values.keys()`.

For plain attribute access, `for smoother in smoother.smoothers:` reads the iterable before the loop target is rebound, and the rebound name does not change the object that supplied `.smoothers`. That makes it closer to a normal attribute traversal than to the self-iteration bug B020 is trying to catch.

The patch keeps the existing method-call behavior (`values.items()`, etc.) because those still indicate iterating over the same named object. But I understand if the project prefers the more conservative warning despite the false positive.

cooperlees:
I don't think we want to smoother, the  'smoother' variable in general so I don't agree it's a false positivehere.

After the loop in this example, isn't smother, smothered?

```python
smother = SmotherObject()
for smoother in smoother.smoothers:
  ...
# smother = the last iterated object in smother.smoothers right? This is bad ...  
assert smother.smoothers  # False
```

Am I missing something here?

binggao1230:
You are right. I focused on the iterable being evaluated before the loop and overlooked the binding afterward: after a non-empty loop, `smoother` is the last child, not the original container. The tuple-unpacking `series` example has the same problem. That is exactly what B020 is meant to flag, so my false-positive premise was wrong and this PR should not proceed. Thanks for catching it.

PR #564, opened 2026-07-14
Add B913 for unused trailing zip values

### LINKED ISSUE #545: Proposed rule: match zip() arguments and unpacked variables

sapunyangkut:
## Summary

- add the optional `B913` diagnostic for direct `zip()` loops with unused or repeated trailing `_` targets
- distinguish real `_` loads from stores and comprehension-local bindings
- document the rule and cover positive, conservative, starred, arity, `strict=`, nested-target, and loop-length cases

Closes #545.

## Behavior and safety

`B913` remains disabled by default. It only checks direct `zip()` calls with a statically known one-to-one top-level target mapping. Calls with keywords (including `strict=`), starred targets or arguments, arity mismatches, non-trailing or nested `_` targets, and all-`_` targets are skipped.

Removing a `zip()` iterable can change the loop length. The diagnostic therefore recommends keeping a descriptive variable when that iterable intentionally controls iteration, and only suggests removal otherwise. Complex nested rebinding is treated conservatively as usage, favoring missed diagnostics over false positives.

## Validation

- `python -m coverage run -m pytest tests/test_bugbear.py`: 81 passed, 1 skipped
- `python -m coverage report -m`: 98% for `bugbear.py`
- `python -m mypy bugbear.py`: success
- `pre-commit run --files bugbear.py README.rst tests/eval_files/b913.py`: isort, Black, flake8, and rstcheck passed
- `git diff --check`: passed

## AI assistance

This pull request was fully AI-assisted and automatically prepared under the GitHub account owner's authorization. Separate AI review passes checked diagnostic correctness, prior review concerns, tests, and security boundaries; their actionable findings were addressed before publication. No issue or pull-request claim comment was posted.

### DIFF
--- a/tests/eval_files/b913.py
+++ b/tests/eval_files/b913.py
@@ -1,0 +1,76 @@
+# OPTIONS: select=["B913"]
+
+# Errors: direct trailing underscore targets whose values are unused.
+for left, right, _ in zip(xs, ys, limits):  # B913: 17
+    consume(left, right)
+
+for [left, right, _] in zip(xs, ys, limits):  # B913: 18
+    consume(left, right)
+
+for left, right, _, _ in zip(xs, ys, limits, extras):  # B913: 17
+    consume(left, right)
+
+for left, _ in zip(xs, limits):  # B913: 10
+    consume(left)
+
+# Stores, nested bindings, and repeated targets do not use every zip value.
+for left, right, _ in zip(xs, ys, limits):  # B913: 17
+    consume(left, right)
+    _ = object()
+
+for left, right, _ in zip(xs, ys, limits):  # B913: 17
+    consume(left, right, [item for _ in xs])
+
+for left, right, _, _ in zip(xs, ys, limits, extras):  # B913: 17
+    consume(left, right, _)
+
+# No errors: the discarded value is used or its iteration semantics are explicit.
+for left, right, _ in zip(xs, ys, limits):
+    consume(left, right, _)
+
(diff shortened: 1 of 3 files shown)


cooperlees (changes requested):
Thanks for the PR. Can you address copilots findings please.

tests/eval_files/b913.py
```diff
@@ -0,0 +1,76 @@
...
+for left, *_rest, _ in zip(xs, ys, limits):
+    consume(left)
+
+for left, right, _ in zip(xs, *iterables):
+    consume(left, right)
+
+for left, right, (value, _) in zip(xs, ys, pairs):
+    consume(left, right, value)
+
+for left, _, right in zip(xs, limits, ys):
```
JelleZijlstra: Why is this not an error? Doesn't seem meaningfully different from the one above where `limits` is last.

cooperlees (approved):
LGTM now, I do agree with Jelle tho, but seems on purpose that it's not, but does confuse me. What are we missing?

PR #567, opened 2026-08-14
B023: don't flag a function that is only called inside the loop

### LINKED ISSUE #380: `B023` False positive
```python
arr = ['a', 'b', 'c', 'd', 'e', 'f']
mydict = {} 
for i in range(0, len(arr)):
    def foo():
        mydict[arr[i]] = 0
    foo()
```
This example is obviously too simple to be realistic - our actual code has more logic `foo()` and passes parameters to it, but is equivalent in terms of everything the rule cares about. We get an error of `Function definition does not bind loop variable 'i'.Flake8(B023)`. This should not happen because `foo` is not a closure. It's never called outside the loop, and execution is never delayed. `i` should always have the correct value.

The reason we aren't immediately doing `mydict[arr[i]] = 0` is because we have other logic that determines the parameters passed to `foo()`.

### LINKED ISSUE #468: B023: false positive for nested function inside loop
The following (admittedly bad) code produces what I think must be a false positive:
```py
for _ in range(10):
   foo = []
   def bar():
      foo.append(42)

   bar()
```
```
test.py:4:7: B023 Function definition does not bind loop variable 'foo'.
```

At the very least the warning is confusing because `foo` is not actually a loop variable.

Using:
```
flake8 --version
7.0.0 (flake8-bugbear: 24.2.6, flake8-builtins: 2.5.0, mccabe: 0.7.0, pycodestyle: 2.11.1, pyflakes: 3.2.0) CPython 3.10.12 on Linux
```

Eljees:
Fixes #468. Fixes #380.

`check_for_b023` warns whenever a function defined in a loop closes over the loop variable. But a function whose every reference is a **direct call in the loop body** cannot outlive the iteration it was defined in, so the value it closes over is the one the author meant:

    def f():
        results = []
        for x in range(3):
            def g():
                return x
            results.append(g())      # called here, never stored
        return results

The existing `safe_functions` notion covered only a fixed set of shapes — `filter`/`map`/`reduce`, a `key=` argument, and `return lambda: x`. This replaces it with the rule you both described on the issues: warn when the name escapes the loop **or** is referenced as anything other than a direct call, and stay silent otherwise. @jakkdl spelled it out on #380 and @cooperlees agreed with it on #468; I've implemented both halves plus one guard of my own — a decorated definition keeps warning, because a decorator can store the function.

### Measured

`python -m pytest`: **79 passed, 2 skipped — before and after**, on `2155484`.

Still flagged (the name escapes):

| shape | |
| --- | --- |
| `[lambda: i for i in range(3)]` | flagged |
| `out.append(lambda: i)` | flagged |
| `return g` from inside the loop | flagged |
| `d[i] = g` | flagged |
| `@staticmethod` above the definition | flagged |

No longer flagged (every reference is a direct call):

| shape | |
| --- | --- |
| `results.append(g())` | silent |
| `print(g())` | silent |

`tests/eval_files/b023.py` gains the new cases in the project's declarative format; the existing expectations are unchanged — the one line that moves in the diff is the same line at a new offset, not a changed expectation.

`black` and `isort` are clean. `rstcheck` reports two unreferenced hyperlink targets in `README.rst` at lines 456 and 463 — that is identical on a clean checkout and unrelated to the change here.

AI-assisted (LLM used for drafting); the runs above are mine.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -501,6 +501,8 @@ UNRELEASED
 ~~~~~~~~~~
 
 * B019: also flag `async_lru.alru_cache` and check cache decorators on `async def` methods (#488)
+* B023: don't flag a function whose every reference is a direct call inside the loop body:
+  such a function cannot outlive the iteration it was defined in (#468, #380)
 * B018: handle also useless calls such as `isinstance(x, int)` without assigning or using the result
 * B031: don't count a store-context reference (e.g. an annotation target like `group: T`) as a use of the `groupby` generator (#465)
 * B902: don't raise a false positive on a metaclass defined with a dotted base such as `abc.ABCMeta` or `enum.EnumMeta` (#411)
--- a/bugbear.py
+++ b/bugbear.py
@@ -1015,6 +1015,7 @@ class BugBearVisitor(ast.NodeVisitor):
         # implement this "backwards": first we find all the candidate variable
         # uses, and then if there are any we check for assignment of those names
         # inside the loop body.
+        immediately_called = self._immediately_called_functions(loop_node)
         safe_functions = []
         suspicious_variables = []
         for node in ast.walk(loop_node):
@@ -1047,6 +1048,15 @@ class BugBearVisitor(ast.NodeVisitor):
             if isinstance(node, ast.Return):
                 if isinstance(node.value, FUNCTION_NODES):
                     safe_functions.append(node.value)
+
+            # a function that is only ever *called* in the loop body cannot
+            # outlive the iteration its free variables were assigned in
+            if (
+                isinstance(node, (ast.AsyncFunctionDef, ast.FunctionDef))
+                and not node.decorator_list
+                and node.name in immediately_called
+            ):
+                safe_functions.append(node)
 
             # find unsafe functions
             if isinstance(node, FUNCTION_NODES) and node not in safe_functions:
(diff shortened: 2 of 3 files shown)


cooperlees (changes requested):
LGTM - Thanks for this.

I think copilot has found some nice things to polish up - Feel free to state why it's wrong tho if it is.

Eljees:
Thanks — the Copilot points hold up, and four of them are the same root cause: I proved "called in this iteration" with `ast.walk`, which is neither scope-aware nor order-aware.

Concretely:

- Deferred bodies (1) is a real hole. `async def` and generator functions return an object instead of running, so the late-binding read survives the call. I will restrict the exemption to synchronous, non-generator definitions.
- Identifier-only keying (2) and calls reached through a nested function (5) are the same defect seen from two sides: a reference has to be tied to the definition that binds it, and matching identifier text alone is the wrong key. I will resolve references against the defining scope.
- `ast.walk(loop_node)` covering the `else` suite and accepting a call that precedes the `def` (4) is correct as well; the search has to be restricted to the statements that can run in the same iteration after the definition.
- The quadratic walk (3) goes away with the same rewrite: one reference/call index built per scope instead of two walks per loop.

I will push a revision with regression cases for each of the five and report back. If you would rather see this as a narrower rule — for example exempting only a direct call in the loop body at statement level — say so and I will cut it to that instead.

cooperlees:
I'm happy to start simpler, but if you're happy to do all five go for it. Both ways are an imrpovement so will take what ever you have time for :) ... thanks!

Eljees:
Went with all five. Also rebased onto main, so the conflict is gone.

The rewrite ties the exemption to the *definition* rather than to the identifier. A name is exempted only when it has at least one reference and every reference in the scope holding the loop is a call that runs in the same iteration as the `def`.

| review point | what changed |
| --- | --- |
| deferred bodies | only a plain `def` qualifies now — an `async def` builds a coroutine and a generator function builds a generator, so the read of the loop variable is deferred |
| identifier-only keying | a name bound anywhere else in the scope is skipped entirely, so an unrelated binding can no longer decide the outcome |
| `ast.walk(loop_node)` too wide | the search covers the loop body only: the `else` suite runs after the loop, and a call above the `def` invokes the previous iteration's binding |
| calls through a nested function | a reference reached through a nested function, lambda or generator expression disqualifies |
| two walks per loop | the parent links, name uses and call targets of a scope are built once and cached |

One case the rewrite surfaced that none of us had listed: a definition **nothing refers to** was being exempted, because "every reference is a call" is vacuously true on an empty set. `f_bad_1` in the existing fixture is exactly that, and it is right to keep reporting it — the name stays bound after the loop, so it can be called later with the last iteration's values. Fixed, and the existing expectation is what caught it.

### Measured

`tests/eval_files/b023.py` now holds 31 expectations, all produced exactly — nothing missing, nothing extra. Four are new, one per unsound case:

- `awaited_later` — an `async def` that is called
- `iterated_later` — a generator function that is called
- `called_above_the_def` — the call sits above the `def`
- `called_in_the_else_suite` — the call sits in the `else` suite

The nested-function case (`wrapper` calling `target`) was already in the fixture from the first round and still reports.

`python -m pytest`: **79 passed, 2 skipped** — identical to the baseline on `838c365`. `black` and `isort` clean on both files. `rstcheck` reports one unreferenced hyperlink target in `README.rst`, and does so on an untouched checkout as well.

If any of the five reads as over-strict to you, say which and I will loosen that one rather than the set.

cooperlees (approved):

PR #568, opened 2026-08-15
B020: don't flag rebinding an attribute of the loop's base object

### LINKED ISSUE #248: Possible B020 false positive with instance attribute
This code:
```python
class MyModel:
    value: int

class A:
    model_instance = MyModel()
    test_suite = [1, 2, 3]

    def method(self):
        for self.model_instance.value in self.test_suite:
            print(self.model_instance.value)
```
results in a `B020`, but it looks like a false positive to me since `self.a` is not one of the iterable values.  Admittedly it is a slightly unusual and ambiguous case since `self` _is_ being modified and the plugin can't know what changes setting `self.first` is making to the object and the value of `self.a`.

The actual use case is a test where a Django model object is created once in the test set up and, in each iteration, the loop modifies model attributes, passes the model to a function, and tests the result.

Eljees:
Fixes #248.

### The false positive

`for self.a in self.b` rebinds the **attribute** `a` — it does not rebind the name `self`.
`check_for_b020` compared the bare base name of the loop target against the names in the
iterable, so `self` matched `self`, and every loop over a sibling attribute of the same
object was reported:

```python
class A:
    test_suite = [1, 2, 3]

    def method(self):
        for self.model_instance.value in self.test_suite:  # B020 (false positive)
            print(self.model_instance.value)
```

`self.model_instance.value` and `self.test_suite` are two different bindings, so this loop
cannot reassign the thing it is iterating.

### The fix

Two changes in `check_for_b020`:

* compare the **whole dotted path** (`self.model_instance.value` against `self.test_suite`)
  rather than the base name, on both the target and the iterable side;
* ignore names that only ever appear in **load** context — those are the base of an
  attribute or subscript target, not something the loop rebinds.

### What still errors

The case the check exists for is unchanged, and there is an eval case pinning it:

```python
for self.test_suite in self.test_suite:  # B020: 12, "self.test_suite"
```

### Checks

* `tests/eval_files/b020.py` gains three cases — two that must now be silent and one that
  must still error. Reverting only `bugbear.py` makes the file fail with exactly the two
  extra `B020`s, so the new cases do pin this fix rather than passing incidentally.
* Full suite on `2155484`: **79 passed, 2 skipped** — same as `main`.
* `pre-commit run --all-files`: isort, black, flake8, rstcheck all pass.
* `README.rst` `UNRELEASED` updated.

@cooperlees — you wrote on the issue that you weren't sure this one could be fixed.
Comparing full dotted paths turned out to be enough; happy to adjust the approach if you'd
rather this stayed a known limitation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -501,6 +501,8 @@ UNRELEASED
 ~~~~~~~~~~
 
 * B019: also flag `async_lru.alru_cache` and check cache decorators on `async def` methods (#488)
+* B020: don't flag `for self.a in self.b`: rebinding an attribute is not rebinding the
+  base name, so two different attributes of the same object are two bindings (#248)
 * B018: handle also useless calls such as `isinstance(x, int)` without assigning or using the result
 * B031: don't count a store-context reference (e.g. an annotation target like `group: T`) as a use of the `groupby` generator (#465)
 * B902: don't raise a false positive on a metaclass defined with a dotted base such as `abc.ABCMeta` or `enum.EnumMeta` (#411)
--- a/bugbear.py
+++ b/bugbear.py
@@ -975,7 +975,7 @@ class BugBearVisitor(ast.NodeVisitor):
                 return
 
     def check_for_b020(self, node: ast.For) -> None:
-        iterset = B020NameFinder()
+        iterset = B020AttributeFinder()
         iterset.visit(node.iter)
         iterset_names = set(iterset.names)
 
@@ -984,11 +984,7 @@ class BugBearVisitor(ast.NodeVisitor):
         # attribute of the same object. Compare the whole dotted path instead.
         candidates: dict[str, ast.expr] = dict(_dotted_targets(node.target))
         if candidates:
-            for sub in ast.walk(node.iter):
-                if isinstance(sub, ast.Attribute):
-                    path = _dotted_name(sub)
-                    if path is not None:
-                        iterset_names.add(path)
+            iterset_names |= iterset.paths
 
         # a name that only ever appears in load context is the *base* of an
         # attribute or subscript target, not something the loop rebinds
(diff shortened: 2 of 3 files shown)


cooperlees (changes requested):
LGTM - Thanks for this.

I think copilot has found some nice performance things especially to polish up here - As always, feel free to state why it's wrong tho if it is.

Eljees:
Thanks. Both are right.

The scope point is the substantive one: I collected dotted paths with `ast.walk`, so a comprehension- or lambda-local binding is treated as the same object as the outer one, and `for obj.value in [obj.value for obj in objects]` regresses. I will collect the paths with a visitor that reuses the comprehension and lambda exclusions `B020NameFinder` already applies, and add both scopes as regression cases.

The stale `Should emit` header is my oversight — line 58 will be listed.

Eljees:
Both done, and the branch is rebased onto main so the conflict is gone.

`B020AttributeFinder` now collects the dotted paths as a subclass of `B020NameFinder`, so it inherits the comprehension and lambda exclusions instead of restating them — that was the part I wanted to avoid duplicating, since the two would drift. The one thing the base class does that had to be mirrored is `visit_Lambda`: it pops the lambda arguments out of `names` after visiting the body, so the subclass drops the paths rooted in those arguments the same way.

### Before / after, same three files

| | previous revision | this revision |
| --- | --- | --- |
| `for obj.value in [obj.value for obj in objects]` | reported | **silent** |
| `for obj.value in map(lambda obj: obj.value, objects)` | reported | **silent** |
| `for obj.value in obj.value` | reported | reported |

Both scopes are now in `tests/eval_files/b020.py` as `ok_comprehension_scope` and `ok_lambda_scope`, with `still_an_error_at_module_level` next to them as the control.

### The header was stale by more than one line

Listing line 58 turned out not to be enough: line 32 was missing from it too, from before this branch. The header now reads

    B020 - on lines 8, 21, 32, 36, 58, and 75

and those are exactly the six the checker produces on the file — I compared the parsed expectations against `BugBearChecker.run()` directly, nothing missing and nothing extra.

`python -m pytest`: **79 passed, 2 skipped**, same as the baseline on `838c365`. `black` and `isort` clean on both changed files.

PR #570, opened 2026-08-19
Extend B019 to alru_cache and async methods

### LINKED ISSUE #488: Consider adding alru_cache to B019
The [async-lru](https://github.com/aio-libs/async-lru) project provides a port for Python's built-in [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) function for [asyncio](https://docs.python.org/3/library/asyncio.html).

The same rules apply here with respect to garbage collection on instance methods. As this package is maintained by [aio-libs](https://github.com/aio-libs) it might be worth considering extending this rule to apply to uses of this decorator.

ChrisJr404:
This extends B019 to cover `async_lru.alru_cache` and cache decorators on `async def` methods, as suggested in #488.

`alru_cache` is the async port of `functools.lru_cache` (maintained by aio-libs), so when it decorates a method it keeps instance references alive in exactly the same way and leads to the same memory leak B019 is meant to catch. While looking into it I noticed B019 only ran on synchronous function definitions, so even a plain `@lru_cache`/`@cache` on an `async def` method slipped through. I added `async_lru.alru_cache` and `alru_cache` to the recognised decorators and wired the check into `visit_AsyncFunctionDef`, so all of these now report on both sync and async methods.

I extended `tests/eval_files/b019.py` with an async class covering `alru_cache`, `async_lru.alru_cache`, their called forms, and `lru_cache`/`cache` on async methods, plus the classmethod/staticmethod negatives to make sure those still don't fire. The README description and change log are updated to match.

Closes #488

### DIFF
--- a/tests/eval_files/b019.py
+++ b/tests/eval_files/b019.py
@@ -1,10 +1,14 @@
 """
 Should emit:
-B019 - on lines 73, 77, 81, 85, 89, 93, 97, 101
+B019 - on each of the cache-decorated instance methods below, both the
+synchronous methods and the `async def` methods (including `alru_cache`).
 """
 
 import functools
 from functools import cache, cached_property, lru_cache
+
+import async_lru
+from async_lru import alru_cache
 
 
 def some_other_cache(): ...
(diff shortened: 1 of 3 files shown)


tests/eval_files/b019.py
```diff
@@ -1,11 +1,15 @@
...
-B019 - on lines 73, 77, 81, 85, 89, 93, 97, 101
+B019 - on each of the cache-decorated instance methods below, both the
+synchronous methods and the `async def` methods (including `alru_cache`).
 """
 
 import functools
 from functools import cache, cached_property, lru_cache
 
+import async_lru
+from async_lru import alru_cache
```
cooperlees, line 11: Does this need to be a dep for the test running? How does this not cause an error or am I forgetting we have some import mocking magic?

cooperlees (approved):
Tests pass and it all looks good to me. So thanks!

PR #575, opened 2026-09-04
B028: flag empty skip_file_prefixes tuples

### LINKED ISSUE #510: Check for empty `skip_file_prefixes` in `B028`
As of https://github.com/PyCQA/flake8-bugbear/pull/503 (cf https://github.com/PyCQA/flake8-bugbear/issues/497) the following emits no lint:

```console
❯ cat example.py
import warnings

warnings.warn("test", skip_file_prefixes=())

❯ uvx --with git+https://github.com/PyCQA/flake8-bugbear flake8 --select B028 check example.py

❯ echo $?
0
```

But it probably should: the stacklevel is only set to at least 2 if there is at least one prefix given. Of course, the user may pass in a reference to some unknown tuple, in which case it's probably better to skip the lint - but if they explicitly pass an empty tuple, then it makes sense to still emit the lint.

On the other hand I understand if this is overly pedantic and unlikely, so feel free to close the issue as unplanned!

Sonike:
Fixes #510

## Summary

- report B028 when `warnings.warn` receives an explicitly empty `skip_file_prefixes=()` tuple
- continue skipping calls whose prefixes are non-empty or cannot be determined statically
- document the exception and add regression coverage

An empty tuple does not cause `warnings.warn` to raise the effective stack level, so treating it like a populated `skip_file_prefixes` value suppresses a useful B028 diagnostic.

## Tests

- `.venv/bin/python -m pytest tests/test_bugbear.py -q` (81 passed)
- `.venv/bin/tox -e py314` (81 passed, 98% coverage)
- `.venv/bin/pre-commit run --all-files` (isort, black, flake8, rstcheck passed)

## Limitations

Dynamic expressions remain intentionally conservative: if the value of `skip_file_prefixes` is not a literal empty tuple, B028 is skipped as before.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -237,7 +237,8 @@ limitations make it difficult.
 **B028**: No explicit stacklevel argument found. The warn method from the warnings module uses a
 stacklevel of 1 by default. This will only show a stack trace for the line on which the warn method is called.
 It is therefore recommended to use a stacklevel of 2 or greater to provide more information to the user.
-The check is skipped when skip_file_prefixes is used.
+The check is skipped when ``skip_file_prefixes`` is used, except for an explicitly empty tuple,
+which does not affect ``stacklevel``.
 
 .. _B029:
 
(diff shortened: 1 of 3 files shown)


cooperlees (approved):

cooperlees:
This is an improvement. Thanks!

PR #576, opened 2026-09-05
B031: allow reusing groups after materialization

### LINKED ISSUE #395: B031 false positive when reassigning the returned generator
Code to reproduce:

```python
from itertools import groupby

for _, gen in groupby([]):
    gen = list(gen)
    print(gen)
```

This wrongly triggers B031 even though the generator is indeed saved to a list  as suggested
in the error message. You can make the error go away by assigning to a new variable so
maybe not worth it to track reassignments?

Sonike:
After `group = list(group)`, B031 currently flags later reads of `group` even though it now holds a reusable list. This recognizes explicit `list(group)` and `tuple(group)` assignments back to the same name, including annotated and chained assignments.

The assignment's right-hand side is checked before the name becomes reusable. Conditional paths are merged conservatively, so a conversion on only one branch or inside a potentially empty nested loop does not suppress genuine repeated-use warnings. Arbitrary calls such as `iter(group)` retain the existing behavior; this does not add alias analysis or infer the return types of helper functions.

Fixes #395.

Validation:

- Confirmed that the initial regression cases fail on the current main branch, reporting three spurious B031 diagnostics, and pass with this change.
- `tox -e py313,py314`: Python 3.13 — 80 passed, 1 version-specific skip; Python 3.14 — 81 passed. Both report 98% coverage.
- `pre-commit run --all-files`: isort, black, flake8, and rstcheck passed.
- Regression coverage includes direct/annotated/chained assignments, both and partial conditional branches, previous consumption, assignment to another name, `iter()`, nested loops, and deferred function bodies.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -500,6 +500,8 @@ Change Log
 UNRELEASED
 ~~~~~~~~~~
 
+* B031: allow reusing a group after assigning ``list(group)`` or ``tuple(group)``
+  back to the same name (#395)
 * B019: also flag `async_lru.alru_cache` and check cache decorators on `async def` methods (#488)
 * B023: don't flag a function whose every reference is a direct call inside the loop body:
   such a function cannot outlive the iteration it was defined in (#468, #380)
--- a/bugbear.py
+++ b/bugbear.py
@@ -1362,6 +1362,27 @@ class BugBearVisitor(ast.NodeVisitor):
             ):
                 self.add_error("B026", starred)
 
+    @staticmethod
+    def _is_b031_group_materialization(node: ast.AST, group_name: str) -> bool:
+        if isinstance(node, ast.Assign):
+            targets = node.targets
+        elif isinstance(node, ast.AnnAssign):
+            targets = [node.target]
+        else:
+            return False
+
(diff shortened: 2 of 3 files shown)


cooperlees (approved):
Agree with copilot. Many thanks for adding this!

cooperlees:
@copilot resolve the merge conflicts on this branch.

Sonike:
Resolved the changelog conflict by merging the latest `main` in a5db492f47f4685045a12182e14a0868c3468b7a, keeping both the B031 and B028 entries. The approved B031 implementation and regression cases are unchanged.

Revalidated the combined result:

- `tox -e py313,py314`: Python 3.13 — 80 passed, 1 version-specific skip; Python 3.14 — 81 passed. Both report 98% coverage.
- `pre-commit run --all-files`: isort, black, flake8, and rstcheck passed.
- Rechecked direct list/tuple, annotated, and chained materialization assignments: four false positives on current `main`, none on this branch; the `iter(group)` reuse warning is retained.

cooperlees (approved):

